PR #9, opened 2022-11-24
Remove `Ground` and `PotentialRef`

alihamdan:
This is the first in a series of PRs in order to simplify the user facing interface of Roseau LF.
In this PR:
* Remove the `Ground` class and usages
* Remove the `PotentialRef` class and usages
* Remove `ElectricalNetwork.special_elements` that hosted the two elements above

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,18 +34,12 @@ en.solve_load_flow(auth=("username", "password"))
 By describing the network and its components, here is a simple example:
 
 ```python
-from roseau.load_flow import Ground, VoltageSource, Bus, PowerLoad, PotentialRef, SimplifiedLine, ElectricalNetwork, LineCharacteristics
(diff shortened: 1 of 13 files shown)


benoit9126:
:warning: J'aimerais comprendre quel est l'objectif poursuivi en retirant ces deux éléments centraux de la modélisation. On en discute dès que vous êtes au Tarmac.

Saelyos:
> @Saelyos there are two error codes `RoseauLoadFlowExceptionCode.NO_POTENTIAL_REFERENCE` and `RoseauLoadFlowExceptionCode.SEVERAL_POTENTIAL_REFERENCE` related to PotentialRef. Are they still needed?

Yes we will need them anyway in `RoseauLoadFlowEngine`, and it's better to have them all in the same place.

PR #10, opened 2022-11-24
Simplify Line classes

alihamdan:
No more `AbstractLine` vs `SimplifiedLine` vs `ShuntLine` but a single `Line` instead. The documentation change will be made in a separate PR.


benoit9126 (approved):
:+1:

PR #11, opened 2022-11-24
Simplify Transformer classes

alihamdan:
Consolidate the transformers in one class. Docs PR to follow.


benoit9126 (approved):
Great again! It will be much better with a single class transformer.

PR #12, opened 2022-11-24
Simplify the Bus classes

alihamdan:
~Here I merge the `AbstractBus`, `VoltageSource`, and `Bus` classes into a single `Bus` class. The former VoltageSource is indicated by a Bus of type `"slack"`. Slack was already used in the conversion to/from json and it is a term used by power system engineers in a load flow analysis. It can be constructed by passing a `source_voltages` that is not None to the bus constructor. The constructor docstring explains what the parameter means in power system terms. I will add an example to the tutorials later as well.~

Simplify the buses classes:
* Now there is only a single `Bus` class to represent a generic electrical bus.
* The `VoltageSource`(s) is(are) now separate element(s) that can be set on the network object
* A bus can have a `VoltageSource` attached, but also a `Load`s and other elements we may add in the future.
* There is no more a bus of type `"slack"`, although a network with a **single** voltage source attached to a bus will make this bus *similar* to a slack bus (at least in the sense that a slack bus has a **fixed complex voltage** and **unlimited power**).

### DIFF
--- a/roseau/load_flow/models/buses.py
+++ b/roseau/load_flow/models/buses.py
@@ -125,7 +125,7 @@ class Bus(Element, JsonMixin):
 
         potentials = data.get("potentials")
 
-        if data["type"] not in ("bus", "bus_neutral"):
+        if data["type"] in ("bus", "bus_neutral"):
             n = 4 if "neutral" in data["type"] else 3
         else:
             msg = f"Bad bus type for bus {data['id']}: {data['type']}"
--- a/roseau/load_flow/io/dict.py
+++ b/roseau/load_flow/io/dict.py
@@ -98,7 +98,7 @@ def network_to_dict(en: "ElectricalNetwork") -> dict[str, Any]:
             if isinstance(element, AbstractLoad):
                 bus_dict["loads"].append(element.to_dict())
             elif isinstance(element, VoltageSource):
-                bus_dict["sources"] = element.to_dict()
+                bus_dict["sources"].append(element.to_dict())
         buses.append(bus_dict)
 
     # Export the branches with their characteristics
--- a/roseau/load_flow/tests/test_electrical_network.py
+++ b/roseau/load_flow/tests/test_electrical_network.py
@@ -5,6 +5,7 @@ import numpy as np
 import pandas as pd
 import pytest
 import requests_mock
+from pandas.testing import assert_frame_equal
 from shapely.geometry import LineString, Point
 
 from roseau.load_flow.exceptions import RoseauLoadFlowException, RoseauLoadFlowExceptionCode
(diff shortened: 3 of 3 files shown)


roseau/load_flow/models/buses.py
```diff
@@ -39,15 +47,40 @@ def __init__(
...
         """
         super().__init__(**kwargs)
         self.id = id
         self.n = n
+        if source_voltages is None:
+            # The bus is a normal bus (non slack): PV bus or PQ bus.
+            self.type = "bus" if n < 4 else "bus_neutral"
+        else:
+            # The bus is a slack bus: The voltage is known and fixed.
+            self.type = "slack"
```
benoit9126, line 72: I don't really like the term "slack". As far as I understand electrical engineers, they use this word to define the main source voltage of a network (generally the largest), and there is only one slack bus per network. Here, we can have as many of this kind of buses in the network. It is also related to the method they use to solve the load flow problem (with a backward-forward, they take the root of the tree, ...) and if they use the bus of the largest generator that is often to ensure a "better convergence"/"better stability", and so on. Nevertheless, no electrical engineers in the G2ELab was able to explain me this in detail…

Let's give to this type a name to describe what it is: "voltage_source_bus".

Another question: There are two ways to define a voltage source:
1. We can define the phase-to-neutral voltages if there are 4 wires connected to this source ("Wye configuration")
2. We can define phase-to-phase voltages if there are only three wires connected to this source ("Delta configuration")
We should consider adding this option (maybe later) and therefore, I would recommend "voltage_source_bus" and "voltage_source_bus_neutral".

benoit9126 (commented):
A single point of questions remains about the word "slack" which seems inappropriate

benoit9126 (approved):
:+1:

roseau/load_flow/models/voltage_sources.py
```diff
@@ -0,0 +1,91 @@
...
+    @ureg.wraps(None, (None, "V"), strict=False)
+    def voltages(self, voltages: Sequence[complex]) -> None:
+        if len(voltages) != self.n - 1:
+            msg = f"Incorrect number of voltages: {len(voltages)} instead of {self.n - 1}"
+            logger.error(msg)
+            raise RoseauLoadFlowException(msg, code=RoseauLoadFlowExceptionCode.BAD_VOLTAGES_SIZE)
+
+        if isinstance(voltages, Quantity):
+            voltages = voltages.m_as("V")
+        self._voltages = voltages
```
Saelyos, line 72: I'm not sure those property are needed anymore: we used to have a `voltage` property because it was a result (for the `AbstractBus`) but now we can read the voltages of the bus attached to the voltage source.

Saelyos (changes requested):
Small changes, otherwise it looks good !

roseau/load_flow/models/voltage_sources.py
Saelyos: No the voltages can be updated between two load flows (for exemple if you want to simulate the tap uncertainty), so the `update_source_voltages` is still needed. It is overridden is `engine` to set internally the new voltages.

Saelyos (approved):

PR #13, opened 2022-11-30
Simplify the loads

alihamdan:
Remove Delta loads, the connection type of the load can be known from its phases.

~Simplify the loads into a single class that represents a generic load described by the equation below.~

~It is not currently possible to use the full functionality of the load model as the load flow engine expects only constant power (formerly `PowerLoad` and `DeltaPowerLoad`) or constant impedance (formerly `ImpedanceLoad` or `DeltaImpedanceLoad`)~

### DIFF
--- a/roseau/load_flow/models/loads/loads.py
+++ b/roseau/load_flow/models/loads/loads.py
@@ -221,14 +221,14 @@ class CurrentLoad(AbstractLoad):
         self.i = self._validate_value(i)
 
     def to_dict(self) -> dict[str, Any]:
-        ya, yb, yc = self.i
+        ia, ib, ic = self.i
         return {
             "id": self.id,
             "phases": "abc" if self.n == 3 else "abcn",
             "currents": {
-                "ya": [ya.real, ya.imag],
-                "yb": [yb.real, yb.imag],
-                "yc": [yc.real, yc.imag],
+                "ia": [ia.real, ia.imag],
+                "ib": [ib.real, ib.imag],
+                "ic": [ic.real, ic.imag],
             },
         }
 
--- a/roseau/load_flow/models/loads/tests/test_loads.py
+++ b/roseau/load_flow/models/loads/tests/test_loads.py
@@ -194,3 +194,60 @@ def test_flexible_load():
     assert load._powers is None  # cannot use load.powers because it is passing None to pint
     load.powers = [100, 100, 100]
     assert (load.powers.m_as("VA") == [100, 100, 100]).all()
+
+
+def test_loads_to_dict():
+    bus = Bus("bus", 4)
+    values = [1 + 2j, 3 + 4j, 5 + 6j]
+
+    # Power load
+    assert PowerLoad("load_s1", 4, bus, values).to_dict() == {
+        "id": "load_s1",
+        "phases": "abcn",
+        "powers": {"sa": [1.0, 2.0], "sb": [3.0, 4.0], "sc": [5.0, 6.0]},
+    }
+    assert PowerLoad("load_s2", 3, bus, values).to_dict() == {
+        "id": "load_s2",
+        "phases": "abc",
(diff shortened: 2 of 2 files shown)


roseau/load_flow/models/loads/loads.py
```diff
@@ -16,562 +15,222 @@
...
-            },
-        }
-
-
-class FlexibleLoad(AbstractLoad):
-    """A class to depict a flexible load i.e. a load with control."""
-
-    flexible_parameter_class: type[FlexibleParameter] = FlexibleParameter
+    def add_control(self, parameters: list[FlexibleParameter]) -> None:
+        """Make the load flexible load i.e. a load with control.
```
Saelyos, line 214: Having a method here to add a control is complicated for the `engine` part, because the internal C++ element should be determined during the constructor, since the `PowerLoad` and the `FlexibleLoad` lead to different equations. I think a `parameters: Optional[list[FlexibleParameter]]` can be added to the constructor of `Load` instead of this method.

roseau/load_flow/models/loads/loads.py
```diff
@@ -16,562 +15,222 @@
...
+    if s is not None and z is not None:
+        raise RoseauLoadFlowException(
+            msg="Load currently expects only s (constant power) or z (constant impedance), not both.",
+            code=RoseauLoadFlowExceptionCode.BAD_LOAD_TYPE,
+        )
+    elif s is None and z is None:
+        raise RoseauLoadFlowException(
+            msg="Load currently expects either s (constant power) or z (constant impedance), non given.",
+            code=RoseauLoadFlowExceptionCode.BAD_LOAD_TYPE,
+        )
```
Saelyos, line 44: I don't understand what are the mixed loads we are talking about, is this something like 30% s, 30% z, 40% i ?

benoit9126: Le plus simple serait de laisser trois types de charge: une à puissance constante, une à intensité constante et une troisième à impédance fixe. L'utilisateur voulant mélanger ces charges n'a qu'à connecter trois charges à un même bus. Cela présente quelques avantages:
1. C'est simple!
2. Si on mélange les charges, on va forcément avoir des demandes du type: "la partie ma charge à puissance constante tire combien de courant?". Réponse : "avec la charge combinée, on n'a que la somme qui est disponible... "

La partie à simplifier de la charge est sur la connexion : je souhaite fusionner les charge en delta et en étoile pour n'avoir qu'une classe par type de charge (puissance, intensité et impédance)

roseau/load_flow/models/loads/loads.py
```diff
@@ -16,562 +15,221 @@
...
+        self.z = self._clean_value(value, type="z")
 
-        for zi in z:
-            if np.isclose(zi, 0):
-                msg = f"An impedance for load {self.id!r} is null"
-                logger.error(msg)
-                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_Z_VALUE)
-        self.z = z
+    def is_flexible(self) -> bool:
+        """Check if the load is flexible.
```
benoit9126, line 214: Could be a property

benoit9126 (changes requested):

benoit9126 (changes requested):
Much easier to read! A small change to avoid confusion

Saelyos:
I'm wondering if merging `PowerLoad` and `FlexibleLoad` by adding `parameters: Optional[list[FlexibleParameter]]` to the constructor wasn't a good idea. Because the `FlexibleLoad` are a kind of `PowerLoad` and  they both have `powers` that can be updated.
What do you think @benoit9126 ?

Saelyos:
Ok, keep whatever you feel is clearer for the user!

benoit9126:
I agree with the proposal to have a "flexible_parameters" args in the constructor of the power load class which is "constant" by default.

Issue #19, opened 2022-12-15
JSON format


benoit9126:
Some modifications to the JSON format:
- [x] Make all objects top-levels. Require #17 
- [x] Add a version number to the file
- [x] Ensure that the ground connections and potentials references connection are well saved and restored. Require #16

Issue #17, opened 2022-12-15
Exit "special elements"


benoit9126:
This issue requires #15 
The name "special elements" is not really meaningful. 
- [ ] Define a "grounds" section in the network and in the JSON export. 
- [ ] Define a "potential_refs" section in the network and in the JSON export.

PR #22, opened 2022-12-20
All elements have id and are json serializable

### LINKED ISSUE #16: Phases management
In order to remove the `n` argument to constructor of classes, we will use phases as strings. Here is the definition of these new arguments:
- [ ] `Bus`: 
  - [ ] the allowed phases are "abc", "abcn", "ab", "bc", "ca", "an", "bn", "cn", "abn", "acn", "bcn". 
  - [ ] This argument of the constructor is mandatory.
- [ ] `VoltageSource`: 
  - [ ] the allowed phases are the same as for the buses. 
  - [ ] Nevertheless, it is not required. By default, the phases of the bus are taken
  - [ ] The phases must be present in the bus it is connected to.
- [ ] `AbstractLoad`: idem as voltage source. NB: When the phase contains a "n", the load will be a star-load, otherwise a delta-load.
- [ ] `Line`:
  - [ ] Only a single `phases` argument to the constructor. The allowed values are "abc", "abcn", "ab", "bc", "ca", "an", "bn", "cn", "abn", "acn", "bcn", "a", "b", "c", "n".
  - [ ] Nevertheless, it is not required. The intersection of the phases of the two buses at the extremities of the line are taken by default.
  - [ ]  The phases of the line must be present in the upstream and in the downstream buses.
  - [ ] The size of the matrices of the line characteristics must be equal to the number of phases of the line.
- [ ] `Switch` ⇾ Same as `Line`
- [ ] `Transformers`
  - [ ] The arguments `phases1` and `phases2` must be added in the constructor of the class.
  - [ ] For the moment, only the phases "abc" and "abcn" are accepted
  - [ ] It is not mandatory to define them. The windings of the transformer characteristics define the default: "D" ⇾ "abc", "Y" ⇾  "abcn"
  - [ ] The `phases1` must be present in the upstream bus.
  - [ ] The `phases2` must be present in the downstream bus.
  - [ ] The winding of the transformer characteristics must be compliant with the selected `phases1` and `phases2`.
- [ ] `Ground`
  - [ ] The constructor only have an `id` argument
  - [ ] The `Ground` class has a method `connect which takes two arguments: a) the bus to connect to; b) the phase to connect to
  - [ ] The `phase` argument of the `connect` method only accepts "a", "b", "c" or "n" as argument.
  - [ ] Its default value is "n".
  - [ ] The chosen phase must be available in the bus.
- [ ] `PotentialRef`
  - [ ] Its constructor takes an element which is a bus or a ground instance. 
  - [ ] Its constructor takes a `phase` argument which can be `a`, `b`, `c` or `n` or `None`.
  - [ ] The default value of the `phase` argument is `None`
  - [ ] If the provided element is a ground, its potential is fixed at 0V. 
  - [ ] If the provided phase is `None` and if the bus' phases contains "n", we fix the potential of the neutral of the bus to zero. Otherwise, we define the sum of potentials of the bus to 0V
  - [ ] If the provided phase is not `None`, we check that this value is present in the bus' phase and we fix it to 0V
  - [ ] In the JSON export, the value provided in the constructor must be serialized (`None` -> `null`)

### LINKED ISSUE #20: Connection simplification
Add a `connect` method to element to avoid the repetition of lines like
```python
self.connected_elements = [bus]
bus.connected_elements.append(self)
```

alihamdan:
Closes #20 (add connect method)
Closes #16 (json export of potential ref phase)
Ref #19 (ground and potential ref in json)

# Changes

* Add ID to all elements via an interface class `Identifiable` (subclassed by `Element`). The ID
  must be a string or an integer. It must be unique among all elements of the same type. The
  per-type uniqueness of ID was already enforced by the `ElectricalNetwork` constructor for most
  elements. The uniqueness of the id of grounds and potential refs will be enforced after their
  liberation from `special_elements` in #17.
* Make all elements json serializable using `JsonMixin` (subclassed by `Element`).
* Add a `connect` method to `Element` to connect to other elements. This method handles the two-way
  connection, ie add the first element to the list of the connected elements of the second and
  vice-versa. It also handles the case where the two elements are already connected.
* Add a `connect_to_bus` method to `Ground` that connects it to a bus at a specific phase. The
  `ground` argument to `Bus` has been removed in favor of this method.
* Removed the confusing single-element results methods on the electrical network such as
  `net.bus_potentials(id)` et al that can be easily confused with `buses_potentials`. These methods
  are equivalent to the equally simple access by element id `net.bus[id].potentials`.
* Make `ElectricalNetwork.remove_element()` take an element instead of an ID. This allows the
  elements to have unique IDs by element type only and not network wide.
* `LineCharacteristics` and `TransformerCharacteristics` have an ID and are json serializable.
* The classes `Control`, `Projection` and `FlexibleParameter` now inherit from `JsonMixin`. They
  already implemented its interface.

### DIFF
--- a/roseau/load_flow/models/core.py
+++ b/roseau/load_flow/models/core.py
@@ -52,7 +52,7 @@ class Element(ABC, Identifiable, JsonMixin):
             logger.error(msg)
             raise RoseauLoadFlowException(msg, RoseauLoadFlowExceptionCode.BAD_PHASE)
 
-    def connect(self, element: "Element") -> None:
+    def _connect(self, element: "Element") -> None:
         """Connect this element to another element.
 
         Args:
@@ -66,6 +66,7 @@ class Element(ABC, Identifiable, JsonMixin):
 
     def disconnect(self) -> None:
         """Remove all the connections with the other elements."""
+        # TODO: make private
         for element in self.connected_elements:
             element.connected_elements.remove(self)
 
@@ -115,7 +116,7 @@ class PotentialRef(Element):
             msg = f"Potential reference {self.id!r} is connected to {element!r} which is not a ground nor a bus."
             logger.error(msg)
             raise RoseauLoadFlowException(msg, RoseauLoadFlowExceptionCode.BAD_ELEMENT_OBJECT)
-        self.connect(element)
+        self._connect(element)
         self.phase = phase
 
     def __repr__(self) -> str:
@@ -159,7 +160,7 @@ class Ground(Element):
 
     1. Connecting to a bus:
 
-       To connect a ground to a bus on a given phase, use the :meth:`Ground.connect_to_bus` method.
+       To connect a ground to a bus on a given phase, use the :meth:`Ground.connect` method.
        This method lets you specify the bus to connect to as well as the phase of the connection.
        If the bus has a neutral and the phase is not specified, the ground will be connected to the
        neutral, otherwise, an error will be raised because the phase is needed.
@@ -186,7 +187,7 @@ class Ground(Element):
     def __repr__(self) -> str:
         return f"{type(self).__name__}(id={self.id!r})"
 
-    def connect_to_bus(self, bus: "Bus", phase: str = "n"):
+    def connect(self, bus: "Bus", phase: str = "n") -> None:
         """Connect the ground to a bus on the given phase.
 
         Args:
@@ -202,7 +203,7 @@ class Ground(Element):
             msg = f"Cannot connect a ground to phase {phase!r} of bus {bus.id!r} that has phases {bus.phases!r}."
             logger.error(msg)
             raise RoseauLoadFlowException(msg, RoseauLoadFlowExceptionCode.BAD_PHASE)
-        self.connect(bus)
+        self._connect(bus)
         self.phases[bus.id] = phase
 
     @classmethod
@@ -276,8 +277,8 @@ class AbstractBranch(Element):
         self._check_phases(id, phases2=phases2)
         self.phases1 = phases1
         self.phases2 = phases2
-        self.connect(bus1)
-        self.connect(bus2)
+        self._connect(bus1)
+        self._connect(bus2)
         self.geometry = geometry
         self._currents = None
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ p_ref = PotentialRef("pref", element=ground)  # A potential reference
 vn = 400 / np.sqrt(3)
 voltages = [vn, vn * np.exp(-2 / 3 * np.pi * 1j), vn * np.exp(2 / 3 * np.pi * 1j)]
 source_bus = Bus(id="source bus", phases="abcn")
-ground.connect_to_bus(source_bus)
+ground.connect(source_bus)
 vs = VoltageSource(id="source", phases="abcn", bus=source_bus, voltages=voltages)
 
 # Create a load bus and a load
--- a/roseau/load_flow/io/dgs.py
+++ b/roseau/load_flow/io/dgs.py
@@ -82,7 +82,7 @@ def network_from_dgs(  # noqa: C901
         source_bus = buses[bus_id]
 
         sources[source_id] = VoltageSource(id=source_id, phases="abcn", bus=source_bus, voltages=voltages)
-        source_bus.connect(ground)
+        source_bus._connect(ground)
 
     # LV loads
     loads: dict[str, AbstractLoad] = {}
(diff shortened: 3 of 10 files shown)


roseau/load_flow/models/lines/line_characteristics.py
```diff
@@ -24,29 +26,29 @@ class LineCharacteristics:
     )
 
     @ureg.wraps(None, (None, None, "ohm/km", "S/km"), strict=False)
-    def __init__(self, type_name: str, z_line: np.ndarray, y_shunt: Optional[np.ndarray] = None):
+    def __init__(self, id: Id, z_line: np.ndarray, y_shunt: Optional[np.ndarray] = None) -> None:
```
benoit9126, line 29: Maybe use a `star` too in the signature of this constructor to ensure keywords arguments

roseau/load_flow/models/core.py
```diff
@@ -107,22 +133,61 @@ def current(self) -> complex:
...
+        """A map of bus id to phase connected to this ground."""
 
     def __repr__(self) -> str:
-        return f"{type(self).__name__}()"
+        return f"{type(self).__name__}(id={self.id!r})"
 
-    def connect(self, bus: "Bus", phase: str = "n"):
-        """Connect the ground to the bus neutral.
+    def connect_to_bus(self, bus: "Bus", phase: str = "n"):
+        """Connect the ground to a bus on the given phase.
```
benoit9126, line 190: I realize that there are several problems here that I did not expect:

1. A `Ground` has a `connect` method and a `connect_to_bus` method. That is confusing…

    *First proposal*: I would suggest you to add a `**kwargs` in the signature of `Element.connect` and then to overload the `connect` method here by declaring a `phase` argument (default to `None`). 

    The element can be 
      * a Bus and phase is required (if phase is None, take "n")
      * a Ground and phase must not be precised.
      * a Line (for the shunt) and phase is not necessary
    You can call the super method inside the implementation of this method.

    ⇾ First proposal seems uselessly complicated...

2. There is no check in the `connect` method. I can do `bus1.connect(bus2)`. It does not really have a meaning. 
    Moreover, in the actual code, you use `self.connect(ground)` in the `Line` constructor while the `ground.connect(self)` would be forbidden without any difference between `connect` and `connect_to_bus`…

Second proposal: The method `connect` of `Element` is renamed into `_connect` to avoid user to use it directly. In the `Ground` class, you rename the `connect_to_bus` into `connect` (public method. Keep the signature of the `connect_to_bus` method ⇾ `(bus, phase="n")`). In this method, use the `_connect` as you have done.

The user will not be able to `connect` elements except a ground to a bus' phase. The `disconnect` method remains public for the moment.


*NB for later:*
Later, the disconnect method may also become protected. We need to think at a clear interface, as there are methods `add_element` and `remove_element` in the `ElectricalNetwork` too... At some point, any element may be aware of the electrical network it belongs to (or if it does not belong to a network). Calling `bus.disconnect(load)` will also trigger something in the electrical network they belong too to ensure that the underlying C++ network model is declared as invalid and the results of all elements of this network are outdated (by logging a warning for instance). It may also ensure that after the call to any `update_*` method, the results are also declared as invalid.

benoit9126 (changes requested):
Good job. Small changes for the moment and a note for later.

roseau/load_flow/models/lines/line_characteristics.py
benoit9126: It may be the same in the `TransformerCharacteristics`

benoit9126: Oh yes, I did not realize.

roseau/load_flow/network/electrical_network.py
```diff
@@ -56,27 +56,31 @@ class ElectricalNetwork:
...
         self,
-        buses: Union[list[Bus], dict[Any, Bus]],
-        branches: Union[list[AbstractBranch], dict[Any, AbstractBranch]],
-        loads: Union[list[AbstractLoad], dict[Any, AbstractLoad]],
-        voltage_sources: Union[list[VoltageSource], dict[Any, VoltageSource]],
+        buses: Union[list[Bus], dict[Id, Bus]],
+        branches: Union[list[AbstractBranch], dict[Id, AbstractBranch]],
+        loads: Union[list[AbstractLoad], dict[Id, AbstractLoad]],
+        voltage_sources: Union[list[VoltageSource], dict[Id, VoltageSource]],
         special_elements: list[Element],
```
benoit9126, line 63: In a next PR, this "special_elements" must be deleted. I think it is the last one.

benoit9126 (approved):

PR #25, opened 2022-12-21
Merge FlexibleLoad into PowerLoad

### LINKED ISSUE #21: Flexible loads
The class `PowerLoad` can be used to model a `FlexibleLoad` with an additional argument in its constructor (`flexible_parameters`, `None` by default). It must be done after #18

alihamdan:
Closes #21 

I also renamed the property `powers` to `flexible_powers` so that it does not shadow the `powers` property we will have later. Later, all of these will have the `res_` prefix.

### DIFF
--- a/roseau/load_flow/models/loads/loads.py
+++ b/roseau/load_flow/models/loads/loads.py
@@ -22,7 +22,7 @@ class AbstractLoad(Element, metaclass=ABCMeta):
     _power_load_class: type["PowerLoad"]
     _current_load_class: type["CurrentLoad"]
     _impedance_load_class: type["ImpedanceLoad"]
-    _flexible_parameter_class: type[FlexibleParameter]
+    _flexible_parameter_class = FlexibleParameter
 
     _type: Literal["power", "current", "impedance"]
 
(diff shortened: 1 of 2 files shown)


roseau/load_flow/models/loads/loads.py
```diff
@@ -305,86 +369,6 @@ def to_dict(self) -> JsonDict:
...
-        self._powers = value
-
-    def to_dict(self) -> JsonDict:
-        return {**super().to_dict(), "parameters": [p.to_dict() for p in self.parameters]}
-
-
 AbstractLoad._power_load_class = PowerLoad
 AbstractLoad._current_load_class = CurrentLoad
 AbstractLoad._impedance_load_class = ImpedanceLoad
-AbstractLoad._flexible_load_class = FlexibleLoad
```
benoit9126, line 390: I think this last line should be replaced with `AbstactLoad._flexible_parameter_class = FlexibleParameter`. (Here or somewhere else)

benoit9126 (approved):
An additional class that disappears! There is a detail I am not sure.

roseau/load_flow/models/loads/loads.py
```diff
@@ -22,7 +22,7 @@ class AbstractLoad(Element, metaclass=ABCMeta):
     _power_load_class: type["PowerLoad"]
     _current_load_class: type["CurrentLoad"]
     _impedance_load_class: type["ImpedanceLoad"]
-    _flexible_load_class: type["FlexibleLoad"]
+    _flexible_parameter_class = FlexibleParameter
```
benoit9126, line 25: We will need to check in January with @Saelyos to see if you can define it here directly. That's sounds good for now.

benoit9126 (approved):

PR #26, opened 2022-12-22
New json format

### LINKED ISSUE #19: JSON format

alihamdan:
Closes #19

## What's changed?
- The produced json now contains the version of the tool as well as basic system information and python environment information for debugging.
- The elements are all represented as top-level lists.
- Since we now have precise phases information for each element, duplicate phase information in the matrices of power, current, or impedance has been removed.


roseau/load_flow/io/dict.py
```diff
@@ -22,6 +27,29 @@
...
+        "platform": sys.platform,  # linux, win32, darwin etc.
+        "machine": platform.machine(),  # x86_64, amd64, i386 etc.
+        # Python environment (useful for debugging)
+        # -----------------------------------------
+        # Only packages that may affect the json data are included
+        "python_version": sys.version_info,
+        "numpy_version": version("numpy"),
+        "pint_version": version("pint"),
+        "shapely_version": version("shapely"),
+    }
```
benoit9126, line 50: I expected something like `{"version": 1}`. My purpose is just to make a check when parsing the JSON in `network_from_json`. I also would like the version of the data to be independent of the version of the software. 

Here, you are implementing a `show_version` function just like in [Pandas](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.show_versions.html) for instance. Good idea, but in another PR.

roseau/load_flow/io/dict.py
```diff
@@ -159,10 +202,13 @@ def network_to_dict(en: "ElectricalNetwork") -> JsonDict:
     transformer_params.sort(key=lambda x: x["id"])  # Always keep the same order
 
     return {
+        "system_info": _get_system_info(),
```
benoit9126, line 205: Just a version field with an integer

roseau/load_flow/models/core.py
```diff
@@ -70,6 +68,15 @@ def disconnect(self) -> None:
         for element in self.connected_elements:
             element.connected_elements.remove(self)
 
+    @staticmethod
+    def _parse_geometry(geometry: Union[str, None, Any]) -> BaseGeometry:
```
benoit9126, line 72: `... -> Optional[BaseGeometry]`

benoit9126 (changes requested):
Two remarks:
- Not so much information in the JSON file. This format is made for exchanging data, not to debug.
- There are in the repository JSON files depicting networks. They need to be upgraded to fit the new file format.

benoit9126:
Yes, exactly. They are used for testing purposes in the engine repository. I think it could be great to have some migrations scripts available in order to automatically upgrade network files. (only for a few old versions. I hope not more than 1 or 2)

```python
from roseau.load_flow import ElectricalNetwork
en=ElectricalNetwork.from_json(<file in version n-1>)
# -> Warning: outdated file. Please save it again. 
# Under the hood, a migration script has been automatically applied. So, no consequences for the user. He can continue to work as usual with the `en` instance

# We would like him/her to do to have a file in version <n> using
en.to_json(<new file>)
```

benoit9126 (approved):
Great! I hope that we will never have more than one version to patch for the users.

PR #36, opened 2023-01-04
Transformer phases update

Saelyos:
Instead of checking if there is a `"n"` in the transformer parameter for the default phases, I prefer to have the intersection between the bus phases and `"abcn"` for `"y/z"` or `"abc"` for `"d"` winding. It seems more coherent with the connections of the other elements of the API (especially the lines) which we could describe as "by default connect everything we can and check that this is coherent with the models we have".

It also seems more natural to me that the phases are **not** determined by a transformer parameter, because a parameter should not define the way the transformer is connected. So with that change, the phases are now determined when the connection is made.

I changed the checks accordingly.

*I'm going to let this PR open for now if I need other changes for `engine`.*

### DIFF
--- a/roseau/load_flow/models/transformers/transformers.py
+++ b/roseau/load_flow/models/transformers/transformers.py
@@ -78,7 +78,9 @@ class Transformer(AbstractBranch):
         if phases1 is None:
             phases1 = "abcn" if w1_has_neutral else "abc"
             phases1 = "".join(p for p in bus1.phases if p in phases1)
+            self._check_phases(id, phases1=phases1)
         else:
+            self._check_phases(id, phases1=phases1)
             # Check that the phases are in the bus
             phases_not_in_bus1 = set(phases1) - set(bus1.phases)
             if phases_not_in_bus1:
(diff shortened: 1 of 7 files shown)


benoit9126 (approved):

roseau/load_flow/models/transformers/transformers.py
```diff
@@ -73,47 +73,58 @@ def __init__(
 
         self.tap = tap
         # Compute the phases if not provided, check them if provided
-        w1_has_neutral = "n" in parameters.winding1.lower()
-        w2_has_neutral = "n" in parameters.winding2.lower()
+        w1_has_neutral = "y" in parameters.winding1.lower() or "z" in parameters.winding1.lower()
+        w2_has_neutral = "y" in parameters.winding2.lower() or "z" in parameters.winding2.lower()
```
alihamdan, line 77: The check was made on the presence of `"n"` because `TransformerType` defines `y` or `z` alone as "without neutral" and `yn` & `zn` as "with a neutral". Doesn't the new check break this assumption?

alihamdan (commented):
Looks good. Just one small question in the comment below.

roseau/load_flow/models/transformers/transformers.py
benoit9126: The good variable name could be `w1_accepts_neutral_point`. If you have a Wye or Zig-zag transformer, there is one.

With this change, we accept that a winding "y" and "yn" are the same. (idem for "z" and "zn"). The connection of the neutral point is fully determined by the `phase1` argument of the constructor (and `phase2` for the second winding). It might lead to a simplification of the `TransformerParameters` class in order to accept `y` and `yn` but keeping internally `"y"` as `winding` field in both case.

alihamdan (approved):

alihamdan:
I updated the json script and the old json files (only 2 affected) to use the same semantics

PR #40, opened 2023-01-17
Floating neutrals in loads and sources do not work yet

alihamdan:
Disallow floating neutrals in loads and sources as they do not work as expected.

### DIFF
--- a/roseau/load_flow/models/voltage_sources.py
+++ b/roseau/load_flow/models/voltage_sources.py
@@ -27,6 +27,7 @@ class VoltageSource(Element):
     """
 
     allowed_phases = Bus.allowed_phases
+    _floating_neutral_allowed: bool = False
 
     def __init__(
         self, id: Id, bus: Bus, *, voltages: Sequence[complex], phases: Optional[str] = None, **kwargs: Any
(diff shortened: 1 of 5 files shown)


benoit9126 (approved):

roseau/load_flow/models/voltage_sources.py
```diff
@@ -58,8 +59,9 @@ def __init__(
             self._check_phases(id, phases=phases)
             # Also check they are in the bus phases
             phases_not_in_bus = set(phases) - set(bus.phases)
-            if phases_not_in_bus and not (phases_not_in_bus == {"n"} and len(phases) > 2):
-                # "n" is allowed to be absent from the bus only if the source has more than 2 phases
+            # "n" is allowed to be absent from the bus only if the load has more than 2 phases
+            floating_neutral = self._floating_neutral_allowed and phases_not_in_bus == {"n"} and len(phases) > 2
+            if phases_not_in_bus and not floating_neutral:
```
Saelyos, line 64: I don't think this is mandatory for `VoltageSource` because we don't have to compute a dissipated power, but we can leave it if we want to have the same interface as the loads.

alihamdan: You are right. The problem is that we currently don't have a way to get the potential of the floating neutral of either the loads or the sources. I changed it until we figure out this floating neutral potential issue.

I don't have a strong opinion here though, if you want to allow it I can change it back.

Saelyos: I don't have a strong opinion either, maybe we can leave it for consistency.

PR #43, opened 2023-01-18
Add a method to get the powers of the loads

alihamdan:
Add a `res_powers` methods to the loads that returns the power in the load elements. The powers here are different from `res_currents` in the sense that `res_currents` always returns the currents in the lines while `res_powers` returns the powers of the element which is Delta powers for Delta loads and Wye powers for Wye loads.

In addition the loads now have `res_voltages` that utilizes the bus potentials to return the voltage across the load elements.

`voltage_phases` is added as a convenience for getting the voltage phases (e.g `["ab"]`, `["an", "bn"]`).

Similar to other `res_` methods, related methods of the loads powers and voltages has been added to the electrical network class to get these quantities for all loads in bulk as a dataframe


roseau/load_flow/models/loads/loads.py
```diff
@@ -85,6 +85,14 @@ def is_flexible(self) -> bool:
...
         return False
 
+    @property
+    def voltage_phases(self) -> list[str]:
+        """The phases of the load voltages."""
+        if "n" in self.phases:  # "an", "bn", "cn"
+            return [p + "n" for p in self.phases[:-1]]
+        else:  # "ab", "bc", "ca"
+            return [p1 + p2 for p1, p2 in zip(self.phases, np.roll(list(self.phases), -1))]
+
```
Saelyos, line 95: I don't think it works for 2-phases delta loads, it should only return `"ab"` and not `"ba"`.

roseau/load_flow/models/voltage_sources.py
```diff
@@ -100,12 +100,19 @@ def voltages(self, voltages: Sequence[complex]) -> None:
...
         self._invalidate_network_results()
 
+    @property
+    def voltage_phases(self) -> list[str]:
+        """The phases of the source voltages."""
+        if "n" in self.phases:  # "an", "bn", "cn"
+            return [p + "n" for p in self.phases[:-1]]
+        else:  # "ab", "bc", "ca"
+            return [p1 + p2 for p1, p2 in zip(self.phases, np.roll(list(self.phases), -1))]
+
```
Saelyos, line 110: Same

Saelyos (changes requested):

Saelyos (approved):

PR #44, opened 2023-01-18
Elements are aware of the network

benoit9126:
Summary of the changes:
- `Element` class has a read only field named `network` which is the network it belongs to (may be `None`)
- The `_connect` method of `Element` now accepts a list of elements and the validity of networks of elements to connect is checked.
- The `disconnect` method of `Element` is now called `_disconnect` and handle the `network`
- The `add_element` method of `ElectricalNetwork` is now called `_connect_element`
- The `remove_element` of `ElectricalNetwork` is now called `_disconnect_element`
- A `disconnect` (public) method has been added to the `VoltageSource` and `AbstractLoad` classes to disconnect them from a network. The disconnected object are not usable anymore after their disconnection, as their field `bus` is set to `None`.
- An user warning is raised when accessing a result from any elements of a network if the network is invalid.
- A (single) user warning is raised when accessing any result of an invalid `ElectricalNetwork`
- Tests have been added

### DIFF
--- a/roseau/load_flow/models/loads/loads.py
+++ b/roseau/load_flow/models/loads/loads.py
@@ -1,5 +1,5 @@
 import logging
-from abc import ABCMeta
+from abc import ABC
 from collections.abc import Sequence
 from typing import Any, Literal, Optional
 
@@ -16,7 +16,7 @@ from roseau.load_flow.utils.units import ureg
 logger = logging.getLogger(__name__)
 
 
-class AbstractLoad(Element, metaclass=ABCMeta):
+class AbstractLoad(Element, ABC):
     """An abstract class of an electric load."""
 
     _power_load_class: type["PowerLoad"]
@@ -70,7 +70,8 @@ class AbstractLoad(Element, metaclass=ABCMeta):
         self._res_currents: Optional[np.ndarray] = None
 
     def __repr__(self) -> str:
-        return f"{type(self).__name__}(id={self.id!r}, phases={self.phases!r}, bus={self.bus.id!r})"
+        bus_id = self.bus.id if self.bus is not None else None
+        return f"{type(self).__name__}(id={self.id!r}, phases={self.phases!r}, bus={bus_id!r})"
 
     @property
     def is_flexible(self) -> bool:
@@ -80,9 +81,7 @@ class AbstractLoad(Element, metaclass=ABCMeta):
     @property
     def res_currents(self) -> np.ndarray:
         """The load flow result of the load currents (A)."""
-        if self._res_currents is None:
-            self._raise_load_flow_not_run()
-        return self._res_currents
+        return self._res_getter(value=self._res_currents, warning=True)
 
     def _validate_value(self, value: Sequence[complex]) -> np.ndarray:
         if isinstance(value, Quantity):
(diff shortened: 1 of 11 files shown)


roseau/load_flow/models/voltage_sources.py
```diff
@@ -100,6 +111,10 @@ def from_dict(cls, data: JsonDict) -> "VoltageSource":
         return cls(data["id"], data["bus"], voltages=voltages, phases=data["phases"])
 
     def to_dict(self) -> JsonDict:
+        if self.bus is None:
+            msg = f"The element {self!r} is disconnected and can not be used anymore."
```
alihamdan, line 115:
```suggestion
            msg = f"The voltage source {self.id!r} is disconnected and can not be used anymore."
```
Same as the loads

roseau/load_flow/models/loads/loads.py
```diff
@@ -223,15 +231,18 @@ def powers(self) -> np.ndarray:
...
     def res_flexible_powers(self) -> np.ndarray:
         """The load flow result of the load flexible powers (VA)."""
-        if self._res_flexible_powers is None:
-            self._raise_load_flow_not_run()
-        return self._res_flexible_powers
+        return self._res_getter(value=self._res_flexible_powers, warning=True)
 
     def to_dict(self) -> JsonDict:
+        if self.bus is None:
+            msg = f"The element {self!r} is disconnected and can not be used anymore."
```
alihamdan, line 243:
```suggestion
            msg = f"The load {self.id!r} is disconnected and can not be used anymore."
```
I think this is a better message for the user but no strong opinion.

roseau/load_flow/models/loads/loads.py
```diff
@@ -295,8 +306,13 @@ def currents(self) -> np.ndarray:
     @ureg.wraps(None, (None, "A"), strict=False)
     def currents(self, value: Sequence[complex]) -> None:
         self._currents = self._validate_value(value)
+        self._invalidate_network()
 
     def to_dict(self) -> JsonDict:
+        if self.bus is None:
+            msg = f"The element {self!r} is disconnected and can not be used anymore."
```
alihamdan, line 313:
```suggestion
            msg = f"The load {self.id!r} is disconnected and can not be used anymore."
```
Same

roseau/load_flow/models/loads/loads.py
```diff
@@ -358,8 +374,13 @@ def impedances(self) -> np.ndarray:
     @ureg.wraps(None, (None, "ohm"), strict=False)
     def impedances(self, impedances: Sequence[complex]) -> None:
         self._impedances = self._validate_value(impedances)
+        self._invalidate_network()
 
     def to_dict(self) -> JsonDict:
+        if self.bus is None:
+            msg = f"The element {self!r} is disconnected and can not be used anymore."
```
alihamdan, line 381:
```suggestion
            msg = f"The load {self.id!r} is disconnected and can not be used anymore."
```

alihamdan (commented):
Looks good. Some minor comments

Saelyos (approved):

alihamdan (approved):

PR #55, opened 2023-02-03
Compatibility with the server

benoit9126:
- Some modifications on the `result_info` field
  - There is no modification of the case (underscore/caml case) performed by the server.
  - Rename some fields in the info on the results. I think that `precision`, `error`, `target_error`, ... may be confusing. For instance `target_error` is `precision`...
  - Rename the field `result_info` into `res_info` (prefixed with `res_` just like all the results)
- I have added a warning message when the server uses a version of `roseau_load_flow` which is greater than the client one.


pyproject.toml
```diff
@@ -50,7 +51,7 @@ requests-mock = "^1.9.3"
 [tool.poetry.group.dev.dependencies]
 coverage = { version = "^7.0.5", extras = ["toml"] }
 flake8 = "^6.0.0"
-pre-commit = "^2.6.0"
+pre-commit = "^3.0.0"
 black = { version = ">=22.10.0", extras = ["jupyter"] }
```
alihamdan, line 55:
```suggestion
black = { version = ">=23.1.0", extras = ["jupyter"] }
```
To keep it in sync with the updated version in `.pre-commit-config.yaml`

roseau/load_flow/network.py
```diff
@@ -324,34 +326,44 @@ def solve_load_flow(
...
+            headers={"accept": "application/json", "rlf-version": __version__},
         )
 
         # Read the response
+        # Check the resonse headers
+        remote_rlf_version = response.headers.get("rlf-new-version")
+        if remote_rlf_version is not None:
+            warnings.warn(
+                message=f"A new version ({remote_rlf_version}) of the library roseau-load-flow is available. Please "
+                f"visit https://github.com/RoseauTechnologies/Roseau_Load_Flow for more information.",
```
alihamdan, line 338: I like the warning, it will be very helpful for the users to see this warning directly when they start experiencing errors due to version differences.
I think it may be better to link to the documentation instead of the repository. I will write a note to remind myself to write a section about this warning and what the user should do in the docs. For now a link to the installation page will be fine.

alihamdan (commented):
Looks good. I like the warning idea. Just left you two small comments.

PR #58, opened 2023-02-13
Warm start argument to `solve_load_flow`

benoit9126:
- Add the `warm_start` argument. It sends the results of the last valid load flow.
- Rename `nb_iterations` to `iterations` to fit the `max_iterations` name.


benoit9126 (commented):
Great!

PR #59, opened 2023-02-14
Improve installations docs and API reference

alihamdan:
* Update the installation docs to use PyPI; mention how to upgrade (TODO: conda upgrade and conda-forge); mention virtual environment as a tip for beginners.
* Improve the API reference
  * Add missing docstrings
  * Add links to used libraries. In functions signatures, the link is automatically created using the type hint (e.g. `def loads_frames(self) -> pd.DataFrame` links to DataFrame docs, etc.)
  * Avoid confusing nesting of the API (e.g. loads are documented as defined in `roseau.load_flow.models` instead of `roseau.load_flow.models.loads.loads`, etc)
* Remove incomplete methods `set_load_point` and `set_source_voltages` (@Saelyos if you need these methods you may want to add them to engine). The tutorials will document how to update the network elements in a more generic way)

### DIFF
--- a/roseau/load_flow/network.py
+++ b/roseau/load_flow/network.py
@@ -1,8 +1,11 @@
+"""
+This module defines the electrical network class.
+"""
 import json
 import logging
 import re
 import warnings
-from collections.abc import Sequence, Sized
+from collections.abc import Sized
 from pathlib import Path
 from typing import NoReturn, Optional, TypeVar, Union
 from urllib.parse import urljoin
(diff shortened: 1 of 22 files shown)


roseau/load_flow/network.py
```diff
@@ -43,6 +46,96 @@
...
+            The potential references of the network. Either a list of potential references or a
+            dictionary of potential references with their IDs as keys. As the name suggests, this
+            element defines the reference of potentials of the network. A potential reference per
+            galvanically isolated section of the network is expected. A potential reference can
+            be connected to a bus or to a ground.
+
+    Attributes:
+        DEFAULT_PRECISION (float):
+            The default precision needed for the convergence of the load flow algorithm. The solver
+            stops when the error between two iterations is below this value. Default is 1e-6.
```
benoit9126, line 92: That's not "between two iterations". At each iteration, we compute the residuals of the equations of the problem. When the maximum of the absolute values of the residuals vector is lower than the provided precision, the iterations stopped.

roseau/load_flow/network.py
```diff
@@ -43,6 +46,96 @@
...
+            Empty if the load flow analysis has not been run yet.
+            Example::
+
+                {
+                    "status": "success",
+                    "resolution_method": "newton",
+                    "iterations": 2,
+                    "target_error": 1e-06,
+                    "final_error": 1.9468870959826745e-12
+                }
```
benoit9126, line 136: Now, there are the additional `warm_start` and `max_iterations` keys. The key `target_error` has been renamed `precision` and `final_error` was renamed `final_precision`

roseau/load_flow/network.py
```diff
@@ -713,7 +818,13 @@ def res_grounds(self) -> pd.DataFrame:
 
     @property
     def res_potential_refs(self) -> pd.DataFrame:
-        """The load flow results of the potential refs as a dataframe."""
+        """The load flow results of the network potential references.
+
+        The results are returned as a dataframe with the following index:
+            - `potential_ref_id`: The id of the potential reference.
+        and the following columns:
+            - `current`: The complex current of the potential reference (in Amps).
```
benoit9126, line 826: If a solution has been found, the current of all potential refs elements should be 0+0j  A.

benoit9126 (approved):
Great. Some minor changes to do.

PR #60, opened 2023-02-15
Units for public methods

benoit9126:
- I have added units to all the properties for data and results. I had a problem with the LineParameters`y_shunt` which can be `None`. A `None` can not be in S/km... Now, I assign a zeros matrix to this field and I have created a method to know if the parameters includes shunt (or not).
- I changed the maintainer in the `__about__` file
- I modified the `JsonMixin` class in order to include the json methods into it. I also dispatched the assignation of results and the results' reader methods.
- I uniformised the `dtype=complex` for numpy arrays. When the project started the behaviour of `dtype=np.complex_` and `dtype=complex` was not the same... It is now the case. Thanks to the Numpy team!


roseau/load_flow/models/lines/lines.py
```diff
@@ -217,10 +218,15 @@ def __init__(
...
-        self.length = length
+        self._length = length
         self.parameters = parameters
         self._initialized = True
 
+    @property
+    @ureg.wraps("km", (None,), strict=False)
+    def length(self) -> Q_:
+        return self._length
+
```
alihamdan, line 229: I think you'll need a setter as well? If you add it you may want to invalidate the results in the setter if the length changes

roseau/load_flow/models/transformers/parameters.py
```diff
@@ -97,15 +98,55 @@ def __eq__(self, other: object) -> bool:
...
+    @property
+    @ureg.wraps("VA", (None,), strict=False)
+    def sn(self) -> Q_:
+        """Losses during off-load test (W)"""
+        return self._sn
+
+    @property
+    @ureg.wraps("W", (None,), strict=False)
+    def p0(self) -> Q_:
+        """Current during off-load test (%)"""
```
alihamdan, line 136: The name is `p0` and the unit is `W`
```suggestion
        """Power during off-load test (%)"""
```

roseau/load_flow/models/transformers/parameters.py
```diff
@@ -97,15 +98,55 @@ def __eq__(self, other: object) -> bool:
...
+
+    @property
+    def i0(self) -> float:
+        """The nominal voltages of the transformer (VA)"""
+        return self._i0
+
+    @property
+    @ureg.wraps("VA", (None,), strict=False)
+    def sn(self) -> Q_:
+        """Losses during off-load test (W)"""
```
alihamdan, line 130:
```suggestion
        """Losses during off-load test (VA)"""
```

roseau/load_flow/models/transformers/parameters.py
```diff
@@ -97,15 +98,55 @@ def __eq__(self, other: object) -> bool:
...
+    @property
+    @ureg.wraps("V", (None,), strict=False)
+    def ulv(self) -> Q_:
+        """Phase-to-phase nominal voltages of the low voltages side (V)"""
+        return self._ulv
+
+    @property
+    def i0(self) -> float:
+        """The nominal voltages of the transformer (VA)"""
+        return self._i0
```
alihamdan, line 125: You forgot the unit wrapper. And the docstring does not seem correct.

roseau/load_flow/models/transformers/parameters.py
```diff
@@ -97,15 +98,55 @@ def __eq__(self, other: object) -> bool:
...
+    @ureg.wraps("W", (None,), strict=False)
+    def psc(self) -> Q_:
+        """Losses during short circuit test (W)"""
+        return self._psc
+
+    @property
+    def vsc(self) -> float:
+        """Voltages on LV side during short circuit test (%)"""
+        return self._vsc
+
```
Saelyos, line 149: The docstring of those methods doesn't mach with the constructor. 
And I think there is a problem in the constructor as well for `sn`, it should be "The nominal power of the transformer"

alihamdan (commented):
Looks good. Just left you some minor comments.

roseau/load_flow/utils/mixins.py
```diff
@@ -26,7 +32,103 @@ def from_dict(cls, data: JsonDict, *args: Any) -> Self:
...
+
     @abstractmethod
     def to_dict(self) -> JsonDict:
         """Return the element information as a dictionary format."""
         raise NotImplementedError
+
+    def to_json(self, path: StrPath) -> Path:
+        """Save the current network to a json file.
+
+        .. note::
```
Saelyos, line 57: Why is this method here and not overridden in `ElectricalNetwork` ? To be called from any element ?

Saelyos (changes requested):

roseau/load_flow/models/transformers/parameters.py
benoit9126: Error in copy-paste. I will correct it.

roseau/load_flow/models/transformers/parameters.py
alihamdan: Yeah you may consider doing something similar to get a uniform interface for all methods

roseau/load_flow/utils/mixins.py
benoit9126: Yes. If you want to have a library of loads, lines, etc. You can export elements to files.

roseau/load_flow/models/transformers/parameters.py
benoit9126: I have added a dimensionless parameter for the `k` (ratio of transformation) of the transformer and for the two unitless electrical arguments of the transformer parameters object. It make sense to transform them into percentage using the new percent unit that I have added. Eg.
```python
k.to("percent") # Q_(0.04, "") ->Q(4, "percent") <=> 4%
```

Saelyos (approved):

alihamdan (approved):

alihamdan:
Thanks.

benoit9126:
I have also looked at `pint pandas` to have units in the resulting data frames, but it seems less finished than `pint` itself. We may reconsider this in the future

PR #61, opened 2023-02-16
Upgrade the test and doc network files

alihamdan:
I ran the upgrade script on the files in `doc/notebooks/data` and did another pass on `roseau/load_flow/tests/data`[^1]. The pass on the test files caught the change to `y_shunt` being always not `None` introduced in #60. I also found that some transformer IDs in the test files were set to `line_...` so I changed them to `tr_...` (also using the script).

[^1]: Some test networks have floating-neutral loads, I had to allow floating neutral in the script otherwise the files could not be parsed by `ElectricalNetwork`


benoit9126 (approved):
Great update!

PR #82, opened 2023-05-05
Add solvers

Saelyos:
Add different solving methods.


benoit9126 (approved):

benoit9126 (changes requested):
@Saelyos I think at Ali's proposal about removing the solver objects from the client library and I think that the best way to do. We definitely need objects on the server side in order to add some methods but, here, a resolution method name and a dictionary of options are sufficient.

Could you please add checks in order to ensure that only valid parameters names are serialized in the request (from the user input dictionary of options)?

benoit9126 (changes requested):
@Saelyos I think at Ali's proposal about removing the solver objects from the client library and I think that the best way to do. We definitely need objects on the server side in order to add some methods but, here, a resolution method name and a dictionary of options are sufficient.

Could you please add checks in order to ensure that only valid parameters names are serialized in the request (from the user input dictionary of options)?

PR #87, opened 2023-05-21
[DOC] Add american standards in LaTeX generated figures

benoit9126:
* Refactor the generation of documentation SVG from LaTeX /TikZ files
* Use the Raleway font in LaTeX generated images
* Use the Raleway font in the HTML documentation
* Use of the new version Font Awesome 6.4.0
* Use of sphinx_inline_tabs to show American vs European standards in electrical schemes
* Rename the iteration number in the `solvers.md` file into `k`. `n` is already the size of the matrix.


alihamdan (approved):
Nice, thanks.

I think we should add diagrams/examples for the Split Phase case as well but we don’t have transformer model for this case yet so this can wait.

PR #88, opened 2023-05-25
Solver update

Saelyos:
- Remove linear solver option, as it was too complicated for the `engine` part. 
- Increase default alpha values for flexible parameters.
- All the solver arguments (`max_iterations`, `tolerance` and `warm_start`) are now passed in the body of the request.

### DIFF
--- a/doc/solvers.md
+++ b/doc/solvers.md
@@ -52,9 +52,7 @@ fails.
 
 ### Parameters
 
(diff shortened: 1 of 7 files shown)


alihamdan (approved):

benoit9126 (approved):
@Saelyos Could you please add in the changelog file a line on the change in the default alpha?

benoit9126:
:+1:

PR #89, opened 2023-05-26
Update dependencies, change Self type

Saelyos:


alihamdan (commented):
Looks good thanks.

We should also add `typing_extensions` to the `intersphinx_mapping` setting in `conf.py` so that sphinx generates links to their page. Something like `"typing_extensions": ("https://typing-extensions.readthedocs.io/en/stable/", None)`.

PR #93, opened 2023-06-08
Short circuits

Saelyos:
- Add short circuit calculation
- Remove absolufy-imports

### DIFF
--- a/roseau/load_flow/models/buses.py
+++ b/roseau/load_flow/models/buses.py
@@ -165,6 +165,8 @@ class Bus(Element):
             phases:
                 The phases to connect.
         """
+        from roseau.load_flow import PowerLoad
+
         for phase in phases:
             if phase not in self.phases:
                 msg = f"Phase {phase!r} is not in the phases {set(self.phases)} of bus {self.id!r}."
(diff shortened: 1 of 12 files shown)


Saelyos:
@benoit9126 
- Removing a short-circuit is not something we can have in the `engine` part, but we can do it here if we want (it will be overridden in `engine` to raise an exception). 
- I'll add that
- I've just added it (we have discussed about it with @alihamdan for the interface)
- I'm not sure what you're referring to, I've added a test with only `"abc"` phases, but otherwise `"n"` is allowed

roseau/load_flow/models/buses.py
```diff
@@ -147,3 +157,25 @@ def _results_to_dict(self, warning: bool) -> JsonDict:
...
+            phases:
+                The phases to connect.
+        """
+        for phase in phases:
+            if phase not in self.phases:
+                msg = f"Phase {phase!r} is not in the phases {set(self.phases)} of bus {self.id!r}."
+                logger.error(msg)
+                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_PHASE)
+        if self._short_circuit is not None:
+            msg = f"A short circuit has already been made on bus {self.id!r}."
```
alihamdan, line 174: Maybe print in the error message the phases that were already short-circuited?

roseau/load_flow/models/buses.py
```diff
@@ -147,3 +157,25 @@ def _results_to_dict(self, warning: bool) -> JsonDict:
             "phases": self.phases,
             "potentials": [[v.real, v.imag] for v in self._res_potentials_getter(warning)],
         }
+
+    def short_circuit(self, *phases: str) -> None:
+        """Make a shot-circuit by connecting multiple phases.
```
alihamdan, line 162:
```suggestion
        """Make a short-circuit by connecting multiple phases together.
```

alihamdan (commented):
Left you some comments

roseau/load_flow/network.py
```diff
@@ -351,6 +351,21 @@ def potential_refs_frame(self) -> pd.DataFrame:
             index="id",
         )
 
+    @property
+    def short_circuits_frame(self) -> pd.DataFrame:
```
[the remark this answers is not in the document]
Saelyos: I've done it in `test_buses` but I'll move it in `test_electrical_network`

benoit9126 (approved):

benoit9126:
@Saelyos Great! Forget what I wrote about the neutral. I read your code too quickly

benoit9126:
@Saelyos Do not forget to add documentation on short circuits.

benoit9126:
@Saelyos  A few additional questions/remarks:
* Could you please add some examples showing that the library protects the user from inconsistent behaviour (in the notebook) ? Example: adding a `PowerLoad` in a bus with a short circuit, etc.
* Currently, it is impossible to model two short circuits in the same bus. Example: I have a ABCN bus and I would like to model a short circuit between the phase A and B and another between the phases C and N. ⇾ It should not be common to do such things, I suppose.

PR #96, opened 2023-07-12
Single-phase and split phase transformers

Saelyos:
- Add support for non-ideal single-phase transformers 
- Add support for non-ideal split-phase transformers 
- Remove `TransformerType`


roseau/load_flow/models/transformers/transformers.py
```diff
@@ -131,39 +174,70 @@ def __init__(
...
-        return {**super().to_dict(), "params_id": self.parameters.id, "tap": self.tap}
+        if phases2 is None:
+            phases2 = "".join(p for p in bus2.phases if p in bus1.phases or p == "n")
+            if len(phases2) != 3:
+                msg = f"Phases (2) of transformer {id!r} can not be deduced from the buses, they need to be specified."
+                logger.error(msg)
+                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_PHASE)
+        else:
+            self._check_phases(id, allowed_phases=self._allowed_phases_split_secondary, phases2=phases2)
+            self._check_bus_phases(id, bus2, phases2=phases2)
```
benoit9126, line 228: In the second option `phases2 is not None`, you ensure that the neutral is present in the secondary phases using the `self._allowed_phases_split_secondary` variable (all the options have a neutral). Not in the first option `phases2 is None`. If I provide `bus1.phases="abcn"` and `bus2.phases="abc"`, I get `phases2="abc"` and `len(phases2)==3`. As far as I remember, the neutral is mandatory for split phase transformer (secondary).

benoit9126 (changes requested):
@Saelyos I think there is a check missing

roseau/load_flow/models/transformers/transformers.py
```diff
@@ -16,19 +16,22 @@
...
 
-    allowed_phases = frozenset({"abc", "abcn"})  # Only these for now
+    allowed_phases = Bus.allowed_phases
     """The allowed phases for a transformer are:
 
-    - P-P-P or P-P-P-N: ``"abc"``, ``"abcn"``
-
-    .. note::
-        Only 3-phase transformers are currently supported.
+    - P-P-P or P-P-P-N: ``"abc"``, ``"abcn"`` (tri-phase transformer)
```
alihamdan, line 27: Let's use "three-phase" in all places visible to the user, you can keep "tri-phase" in the private functions.
```suggestion
    - P-P-P or P-P-P-N: ``"abc"``, ``"abcn"`` (three-phase transformer)
```

roseau/load_flow/models/transformers/transformers.py
```diff
@@ -16,19 +16,22 @@
...
+    - P-P or P-N: ``"ab"``, ``"bc"``, ``"ca"``, ``"an"``, ``"bn"``, ``"cn"`` (single-phase transformer or primary of
+    split-phase transformer)
```
alihamdan, line 29: Keep the indentation correct for a markdown list
```suggestion
    - P-P or P-N: ``"ab"``, ``"bc"``, ``"ca"``, ``"an"``, ``"bn"``, ``"cn"`` (single-phase
      transformer or primary of split-phase transformer)
```

roseau/load_flow/models/transformers/transformers.py
Saelyos: Indeed, good catch. The check with `len` was ok for single-phase, but not here.

roseau/load_flow/models/transformers/transformers.py
```diff
@@ -131,39 +174,70 @@ def __init__(
...
+                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_PHASE)
+        else:
+            self._check_phases(id, allowed_phases=self._allowed_phases_single, phases1=phases1)
+            self._check_bus_phases(id, bus1, phases1=phases1)
 
-    def to_dict(self) -> JsonDict:
-        return {**super().to_dict(), "params_id": self.parameters.id, "tap": self.tap}
+        if phases2 is None:
+            phases2 = "".join(p for p in bus2.phases if p in bus1.phases or p == "n")
+            if len(phases2) != 3:
```
alihamdan, line 222:
```suggestion
            if len(phases2) != 3 or "n" not in phases2:
```

alihamdan (changes requested):
Thanks. I left a few comments, mostly nits.

Could you please replace all "can not" in the error messages by "cannot" (no space)? thanks.

roseau/load_flow/models/transformers/transformers.py
Saelyos: I've used `_allowed_phases_split_secondary` to make it clearer (and replaced all the `len` checks above as well)

roseau/load_flow/models/tests/test_phases.py
```diff
@@ -217,11 +217,11 @@ def test_switches_phases():
     assert switch.phases == switch.phases1 == switch.phases2 == "ca"
 
 
-def test_transformer_phases():
+def test_transformer_tri_phases():
```
benoit9126, line 220: three here too

benoit9126 (approved):

alihamdan (approved):

Issue #69, opened 2023-02-22
DOC: Models section


benoit9126:
A "Models" section is missing in the documentation. Currently, each model has its equations in the API section only. A detailed page per model should be added:
- [x] `VoltageSource`
- [x] Loads: `PowerLoad`, `CurrentLoad`, `ImpedanceLoad`
- [x] Flexibility in power loads
- [x] Transformers
- [x] Lines/Switch
- [x] `Ground`
- [x] `PotentialRef`

For each model, present the underlying equations and a minimal example of usage. If possible, give a link to some example networks which use this model (or to an existing tutorial).

For the flexible power loads, create a simple example in which the effect of the projection can be seen. A single line, a single flexible load and we plot the theoretical control function and the real control function (different because of a projection)

PR #97, opened 2023-07-12
Documentation

benoit9126:
Improvement of the documentation. Creation of a `models` section #69.


benoit9126:
@alihamdan @Saelyos I have improved what I could. Now, there are remaining TODO (single-phase transformer, split-phase transformer, Switch example, etc) that remain in the documentation. Could you please read it when you have time?

alihamdan:
@benoit9126 this is ready now.

We have models docs and usage examples for all elements.

I also fixed all the links 🤞🏻 (the relative link hack you used earlier in the code docstring does not always work, it depends on the page you are in when you click the link, for example, works from `roseau.load_flow.models.line` but not from `roseau.load_flow.line`. Using `:doc:` instead always works 🍾 )

benoit9126:
I will add some other modifications in this branch (Google Analytics, etc). @alihamdan @Saelyos Thanks! :+1:

Small question when reading the documentation. In such example, I think there is a problem with delta connected voltage source (extracted from the example on the page "three-phase transformer"):

```python
# Create a MV bus
bus_mv = Bus(id="bus_mv", phases="abc")

# ....

# Create a voltage source and connect it to the MV bus
voltages = 20e3 * np.exp([0, -2j * np.pi / 3, 2j * np.pi / 3])
vs = VoltageSource(id="vs", bus=bus_mv, voltages=voltages)
```

As it is a delta voltage source, its voltages should have been:

```python
import numpy as np
from roseau.load_flow.converters import calculate_voltages

potentials = 20e3/np.sqrt(3) * np.exp([0, -2j * np.pi / 3, 2j * np.pi / 3])
voltages = calculate_voltages(potentials, "abc")
# array([ 17320.50807569+10000.j,      0.        -20000.j,  -17320.50807569+10000.j])
```

alihamdan:
I think both are correct. It depends on what you'd like to be your "reference" (angle 0 with the potential ref). At the end, voltage is relative and the results you get will depend on your definition but both are correct.

> Small question when reading the documentation. In such example, I think there is a problem with delta connected voltage source (extracted from the example on the page "three-phase transformer"):
> 
> ```python
> # Create a MV bus
> bus_mv = Bus(id="bus_mv", phases="abc")
> 
> # ....
> 
> # Create a voltage source and connect it to the MV bus
> voltages = 20e3 * np.exp([0, -2j * np.pi / 3, 2j * np.pi / 3])
> vs = VoltageSource(id="vs", bus=bus_mv, voltages=voltages)
> ```

In this example, the voltage `Uab` has angle 0 with the potential ref

> As it is a delta voltage source, its voltages should have been:
> 
> ```python
> import numpy as np
> from roseau.load_flow.converters import calculate_voltages
> 
> potentials = 20e3/np.sqrt(3) * np.exp([0, -2j * np.pi / 3, 2j * np.pi / 3])
> voltages = calculate_voltages(potentials, "abc")
> # array([ 17320.50807569+10000.j,      0.        -20000.j,  -17320.50807569+10000.j])
> ```

While in this example `Va` has the angle 0 with the potential ref.

In both cases you get a 20kV source with 120° between the phases.

PR #105, opened 2023-07-28
Make internal class attributes private

alihamdan:
We already did this for the load classes but forgot the others on the network and flexible parameters.


benoit9126 (approved):
:+1: It should also be changed in the engine.

alihamdan:
> +1 It should also be changed in the engine.

I have a commit ready, just waiting for this to be merged

PR #112, opened 2023-08-07
Make the geometry serialization optional

Saelyos:


alihamdan (changes requested):
Thanks. I think it is better to use something like `include_geometry: bool = True` instead to avoid confusion with the `geometry` used for the point or line objects.

Saelyos:
It was inspired by pandas `df.to_json(index=True)`, but I can change it.

benoit9126:
I agree with @alihamdan. I would prefer `include_geometry`

benoit9126 (approved):
@Saelyos Could you add a line a the changelog?

alihamdan (approved):
Looks good. Just a minor comment.

PR #118, opened 2023-08-11
Add a catalogue of transformers

benoit9126:
I have added a catalogue of transformers to the class `TransformerParameters`. I also removed the class method `from_name` of this class.


Saelyos:
Suggestion: in addition the displayed column in `print_catalogue`, add an `id` column which contains names such as `SE_Minera_A0Ak_50`. 

So when the user has chosen a transformer, he can juste copy paste the id:
```python
tp = TransformerParameter.from_catalogue(id="SE_Minera_A0Ak_50")
```
Instead of having to rewrite columns 
```python
TransformerParameters.from_catalogue(efficiency="A0Ak", type=r"^y.*$")
```

benoit9126:
@Saelyos I can add it. The main drawback is that I am not sure that the names will remain the same in the future. I may add a small warning in the documentation on this instability.

alihamdan:
I agree with @Saelyos. If the names change in the future we can provide aliases for backward compatibility.

alihamdan (approved):

PR #138, opened 2023-10-06
Add network constraints for load flow analysis

alihamdan:
The constraints are not used by the load flow solver. They are only used to analyze the results.

The constraints that can be defined include:
- For buses: minimum and maximum voltages. Use `bus.res_violated` to see if the bus has over- or under-voltage.
- For lines: a maximum current. Use `line.res_violated` to see if the loading of any of the line's cables is too high.
- For transformers: a maximum power. Use `transformer.res_violated` to see if the transformer loading is too high.
- The new fields also appear in the data frames of the network.

I still need to change the documentation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,3 +38,4 @@ repos:
     hooks:
       - id: prettier
         args: ["--print-width", "120"]
+        exclude: ^.vscode/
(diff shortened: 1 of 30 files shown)


benoit9126 (approved):
:+1: The documentation remains and it will be good!

benoit9126 (approved):

PR #141, opened 2023-10-30
Add support for Python 3.12 and networkx graphs

### LINKED ISSUE #137: ENH: NetworkX
Add a method to export a network to a [NetworkX](https://networkx.org/documentation/stable/index.html) directed graph. This dependency can be an extra dependency as matplotlib.

alihamdan:
Closes #137
- Add support for Python 3.12
- Add an `ElectricalNetwork.to_graph` method to get a `networx.Graph` object representing the network
- Add an `ElectricalNetwork.buses_clusters` property to group buses on the same voltage levels (i.e. connected via lines or switches)
- Add a `Bus.find_neighbors()` method to get an iterator of buses connected to this bus via a line or a switch
- Fix sphinx builds failing with old autoapi extension version
- Fix missing links to matplotlib in the docs
- Fix missing `plot` extra to install matplotlib. It was misconfigured in pyproject.toml

I initially planned to remove support for Python 3.9 in this release. This will have to wait to the next release as some interested people I met are still using python 3.9 and this release fixes a lot of pandas warnings for them.

### DIFF
--- a/roseau/load_flow/models/buses.py
+++ b/roseau/load_flow/models/buses.py
@@ -191,15 +191,15 @@ class Bus(Element):
             return float(min(voltages)) < self._min_voltage or float(max(voltages)) > self._max_voltage
 
     def propagate_limits(self, force: bool = False) -> None:
-        """Propagate the voltage limits to neighbor buses.
+        """Propagate the voltage limits to galvanically connected buses.
 
-        Neighbor buses here refers to buses connected to this bus through lines or switches. This
+        Galvanically connected buses are buses connected to this bus through lines or switches. This
         ensures that these voltage limits are only applied to buses with the same voltage level. If
         a bus is connected to this bus through a transformer, the voltage limits are not propagated
         to that bus.
 
         If this bus does not define any voltage limits, calling this method will unset the limits
-        of the neighbor buses.
+        of the connected buses.
 
         Args:
             force:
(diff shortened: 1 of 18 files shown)


roseau/load_flow/models/buses.py
```diff
@@ -253,6 +253,29 @@ def propagate_limits(self, force: bool = False) -> None:
             bus._min_voltage = self._min_voltage
             bus._max_voltage = self._max_voltage
 
+    def find_neighbors(self) -> Iterator[Id]:
+        """Find the buses connected to this bus via a line or switch recursively."""
+        from roseau.load_flow.models.lines import Line, Switch
+
```
Saelyos, line 259: I'm wondering if "neighbors" is not a bit confusing. When I first saw this method, I thought it would only give the adjacent buses connected to that bus.

alihamdan: Right. I was going with `get_connnected_buses` at first but it suffers from the same ambiguity as `find_neighbors` in that it could mean buses connected with one line or switch. Do you have a better name? I can also clarify the docstring to say something like `Find all buses connected to this bus via one or more lines or switches, stop at transformers.`

Saelyos: I think we've used the term `galvanic component` for this kind of stuff, but I'm not convinced that it would be understood easily. Between `get_connected_buses` and `find_neighbors`, I think I prefer the former as I find it less confusing, but I agree that there is still an ambiguity that has to be clarified in the docstring.

alihamdan: OK I renamed it to `get_connected_buses` and clarified that we are talking about "galvanically connected" buses.

benoit9126 (approved):

PR #142, opened 2023-10-30
Add voltage unbalance calculation

### LINKED ISSUE #136: ENH: Voltage Unbalance Factor (VUF)
Add a property to get the Voltage Unbalance Factor (VUF) defined as the ratio of the negative sequence voltage phasors by the positive one. Add it in the data frame of buses' results.

See [this paper](https://molzahn.github.io/pubs/girigoudar_molzahn_roald-2019.pdf) for ore details on the subject

alihamdan:
Closes #136

Add method `res_voltage_unbalance` to 3-phase buses that calculates the voltage unbalance factor. Note that this is a method, not a property, so that we can support more standards in the future by passing arguments to this method.

Voltage unbalance does not make sense for non 3-phase buses thus we raise an error with a helpful message.

I also fixed the symmetrical to phasor converters to preserve the shape of the input array.

### DIFF
--- a/roseau/load_flow/models/buses.py
+++ b/roseau/load_flow/models/buses.py
@@ -7,7 +7,7 @@ import pandas as pd
 from shapely import Point
 from typing_extensions import Self
 
-from roseau.load_flow.converters import calculate_voltage_phases, calculate_voltages
+from roseau.load_flow.converters import calculate_voltage_phases, calculate_voltages, phasor_to_sym
 from roseau.load_flow.exceptions import RoseauLoadFlowException, RoseauLoadFlowExceptionCode
 from roseau.load_flow.models.core import Element
 from roseau.load_flow.typing import Id, JsonDict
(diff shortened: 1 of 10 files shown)


roseau/load_flow/models/buses.py
```diff
@@ -279,6 +279,28 @@ def get_connected_buses(self) -> Iterator[Id]:
...
+        if self.phases not in {"abc", "abcn"}:
+            msg = f"Voltage unbalance is only available for 3-phases buses, bus {self.id!r} has phases {self.phases!r}"
+            logger.error(msg)
+            raise RoseauLoadFlowException(msg, code=RoseauLoadFlowExceptionCode.BAD_PHASE)
+        # We use the potentials here which is equivalent to using the "line to neutral" voltages as
+        # defined by the standard. The standard also has this note:
+        # NOTE 1 Phase-to-phase voltages may also be used instead of line to neutral voltages.
+        potentials = self._res_potentials_getter(warning=True)
+        _, vp, vn = phasor_to_sym(potentials[:3])  # (0, +, -)
+        return abs(vn) / abs(vp) * 100
```
benoit9126, line 302: In the `TransformerParameter` class, for the  `__init__` method, the arguments `i0` and `vsc` are percentages too. They are provided as values between 0 and 1 (eg. "0.1" for "10%"). I think we should do the same. The unit `""` is used with `pint` to avoid confusion. If you prefer returning percentages (between 0 and 100), please use the `"percent"` unit of pint.

benoit9126 (changes requested):
A comment on the unit of the result which could be added to avoid confusion.

roseau/load_flow/models/buses.py
alihamdan: Mmmm. On one hand the standars defines VUF as a value between 0 and 100, on the other hand I like to be consistent in the code. It is a tough call.

benoit9126 (approved):

PR #151, opened 2023-12-01
Update to Python 3.10

alihamdan:
I extracted this part from my incoming PR that addresses line parameters. It basically sets the minimum supported Python version at 3.10 and applies the ruff fixes

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9"]
+        python-version: ["3.11"]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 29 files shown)


.github/workflows/conda.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9"]
+        python-version: ["3.11"]
```
benoit9126, line 20: I think you can try python 3.12

benoit9126 (approved):

.github/workflows/conda.yml
alihamdan:
```suggestion
        python-version: ["3.12"]
```

PR #164, opened 2024-01-12
Moving away from SaaS, step 1

alihamdan:
This is the first step towards having the standalone software by abondoning SaaS.

### Remaining Tasks
- [ ] Add changelog entries
- [ ] Add the solver's wheel as a dependency in pyproject.toml (when it becomes available)
- [ ] Do another pass on the docs to make sure they are up-to-date
- [ ] Rerun the skipped CI

### DIFF
--- a/roseau/load_flow/utils/_versions.py
+++ b/roseau/load_flow/utils/_versions.py
@@ -18,7 +18,20 @@ def _get_sys_info() -> JsonDict:
 
 def _get_dependency_info() -> JsonDict:
     """Get versions of dependencies."""
-    return {dist: version(dist) for dist in ("pandas", "numpy", "geopandas", "shapely", "regex", "pint")}
+    return {
+        dist: version(dist)
+        for dist in (
+            "pandas",
+            "numpy",
+            "geopandas",
+            "shapely",
+            "regex",
+            "pint",
+            "platformdirs",
+            "certifi",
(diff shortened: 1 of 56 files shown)


roseau/load_flow/utils/_versions.py
```diff
@@ -18,7 +18,7 @@ def _get_sys_info() -> JsonDict:
 
 def _get_dependency_info() -> JsonDict:
     """Get versions of dependencies."""
-    return {dist: version(dist) for dist in ("pandas", "numpy", "geopandas", "shapely", "regex", "pint", "requests")}
+    return {dist: version(dist) for dist in ("pandas", "numpy", "geopandas", "shapely", "regex", "pint")}
```
alihamdan, line 21: Note to self, add platformdirs and certifi

roseau/load_flow/utils/log.py
```diff
@@ -0,0 +1,85 @@
...
+This is seaborn's default color palette. Generated with:
+```python
+import seaborn as sns
+sns.set_theme()
+list(sns.color_palette().as_hex())
+```
+"""
+
+
+def set_logging_config(verbosity: str):
```
alihamdan, line 46: Note to self. Only keep `cy_set_logging_config` and delete everything else.

PR #165, opened 2024-01-15
Moving away from SaaS, step 2

Saelyos:
- Use license error code
- Improve performance of network creation and results


alihamdan (approved):
Thanks.
I tested the error code locally and now it shows the correct error for both invalid license and bad jacobian.

PR #178, opened 2024-02-07
Merge and deprecate results serialization methods

### LINKED ISSUE #176: Merging `to_dict` and `results_to_dict` methods
I am working on merging the `to_dict` and `results_to_dict` methods (also `..._to_json` methods) and I have a question about the new design.
The current in progress work adds a `include_results=True` parameter to these methods. This means that the results will be included by default when converting to a dict or to json. This however would fail if the results are invalid or if no load flow is run yet. Another option is to make the default `False` and error when the user set it to `True` when results are invalid. I see four options to move forward:

1. Include the results by default (`include_results=True`) and ignore them if they are invalid. Warn the user when trying to convert a network with invalid results.
2. Include the results by default (`include_results=True`) and ignore them if they are invalid silently (no warning).
3. Include the results by default (`include_results=True`) and raise an error when trying to convert a network with invalid results.
4. Do not include the results by default (`include_results=False`). Raise an error when trying to convert a network with explicit `include_results=True` passed by the user and invalid results.

I tend to prefer option 1. I think including the results is a good default and I don't like raising an error on freshly created networks with no load flow results.

What do you think would be the best option?

alihamdan:
Resolves #176

Instead of calling `results_to_dict`, `results_to_json`, `results_from_dict`, and `results_from_json` separately, these are now part of `to_dict`, etc. with an opt out option. The old methods are deprecated for one release and then will be removed. I updated the docs in the Getting Started tutorial with the new usage (I also took advantage of editing that page to make some improvements).


roseau/load_flow/tests/test_electrical_network.py
```diff
@@ -31,147 +30,80 @@
...
+    # line = Line("line", bus0, bus1, phases=phases, parameters=lp, length=1.0, geometry=line_string)
+
+    # return ElectricalNetwork(
+    #     buses=[bus0, bus1],
+    #     branches=[line],
+    #     loads=[load],
+    #     sources=[vs],
+    #     grounds=[ground],
+    #     potential_refs=[pref],
+    # )
```
benoit9126, line 105: You can delete the commented code in the fixture if you want

benoit9126 (approved):
Great! :tada:

Saelyos (approved):

roseau/load_flow/tests/test_electrical_network.py
alihamdan: Moved them to a script

.pre-commit-config.yaml
```diff
@@ -34,3 +34,4 @@ repos:
     hooks:
       - id: prettier
         args: ["--print-width", "120"]
+        require_serial: true
```
alihamdan, line 37: @benoit9126 this turned out to be the reason prettier was very slow in pre-commit.
`pre-commit run prettier -a` now runs in about 1.2 seconds compared to 8 -> 16 seconds without this option.

PR #179, opened 2024-02-08
Bug in `propagate_potentials` with "center" transformer types

benoit9126:
In case of "center" transformers, if the bus of first winding only have two phases (without neutral), the propagation of the potentials fails because three values are expected at the bus of the secondary winding.

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -7,11 +7,13 @@ Starting with version 0.7.0, Roseau Load Flow will no longer be supplied as a Sa
 a standalone Python library.
 ```
 
+- {gh-pr}`179` Solve a bug in the propagation of potentials when a center-tapped transformer is used without neutral at
+  the primary side.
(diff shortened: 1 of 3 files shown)


doc/Changelog.md
```diff
@@ -7,11 +7,13 @@ Starting with version 0.7.0, Roseau Load Flow will no longer be supplied as a Sa
 a standalone Python library.
 ```
 
+- {gh-pr}`179` Solve a bug in the propagation of potentials when a center-tapped transformer is used without neutral at
```
alihamdan, line 10:
```suggestion
- {gh-pr}`179` Fix a bug in the propagation of potentials when a center-tapped transformer is used without neutral at
```

alihamdan (commented):
Thanks. Just a small nit.

Saelyos (approved):

Issue #174, opened 2024-02-05
BUG: JSON export with LineParameters created from catalogue


YassineAbdelouadoud:
**Describe the bug**
The `to_json` method fails when `LineParameters` created using the `from_catalogue` method is present in the network.

`TypeError: Object of type int64 is not JSON serializable`

the problem comes from the max_current and section parameters

**To Reproduce**

```python
from roseau.load_flow import *

  line_parameters = LineParameters.from_catalogue(
      "T_AL_70",
  )
  en = ElectricalNetwork.from_catalogue(name="MVFeeder210", load_point_name="Winter")
  list(en.branches.values())[1].parameters = line_parameters
  en.to_json('tmp.json')
```

**Expected behavior**
Working JSON export

```
System Information
------------------
python               3.10.8.final.0
os                   linux
os_name              posix
machine              x86_64
Installed Dependencies
----------------------
pandas               2.1.0
numpy                1.23.5
geopandas            0.13.2
shapely              2.0.1
regex                2022.10.31
pint                 0.22
platformdirs         4.1.0
certifi              2023.7.22
roseau-load-flow-engine 0.12.0a0

```

PR #168, opened 2024-01-22
Potentials propagation

Saelyos:
- Fix potentials propagation : the correct potentials are now propagated to the LV side (fix #166)
- Allow all elements to be connected after the network creation

PR #167, opened 2024-01-22
Add lines catalogue and get_catalogue method

alihamdan:
Resolves #161
Fixes some of the points of #122

PR #180, opened 2024-02-08
Version 0.7.0

benoit9126:
> [!IMPORTANT]
> Starting with this version, Roseau Load Flow is no longer supplied as a SaaS. The software is
> available as a standalone Python library. Please contact us at contact@roseautechnologies.com in order to get a license key.

This release includes:
- Fix a bug in the engine: it was impossible to change the parameters of center-tapped and single phase transformers.
- #179 Fix a bug in the propagation of potentials when a center-tapped transformer is used without neutral at the primary side.
- #178 #176 Merge the `results_to_json`, `results_from_json`, `results_to_dict`  and `results_from_dict` methods of the `ElectricalNetwork` and `Element`s classes into the methods  `to_json`, `from_json`, `to_dict` and `from_dict` respectively. The old `results_` methods are  **deprecated** and will be removed in a future release. The new methods will include the results by  default, but you can pass `include_results=False` to exclude them.
- #175 #174 Fix JSON serialization of network with line parameters created from the catalogue.
- #173 Remove the conda installation option.

It also includes all the modifications of the alpha release #171 :

- #168 #166 Fix initial potentials' propagation.
- #167 #161 Add a catalogue of lines using the IEC standards. You can use the method `LineParameters.get_catalogue()` to get a data frame of the available lines and the method `LineParameters.from_catalogue()` to create a line from the catalogue. Several line types, conductor material, and insulation types have been updated. Physical constants have been updated to match the IEC standards where applicable.
- #167 The class `LineParameters` now takes optional arguments `line_type`, `conductor_type`, `insulator_type` and `section`. These parameters are accessible as properties. They are filled automatically when creating a line from the catalogue or from a geometry.
- #167 Replace all `print_catalogue()` methods by `get_catalogue()` methods that return a data frame instead of printing the catalogue to the console.
- #167 Enumeration classes no longer have a `from_string` method, you can call the enumeration class directly with the string value to get the corresponding enumeration member. Case-insensitive behavior is preserved.
- #167 #122 Add checks on line height and diameter in the `LineParameters.from_geometry()` alternative constructor. This method will try to guess a default conductor and insulation type if `None` is provided.
- #163 **BREAKING CHANGE:** roseau-load-flow is no longer a SaaS project. Starting with version 0.7.0, the software is distributed as a standalone Python package. You need a license to use it for commercial purposes. See the documentation for more details. This comes with a huge performance improvement but requires a breaking change to the API:
  - The `ElectricalNetwork.solve_load_flow()` method no longer takes an `auth` argument.
  - To activate the license, you need to call `roseau.load_flow.activate_license("MY LICENSE KEY")` or set the environment variable `ROSEAU_LOAD_FLOW_LICENSE_KEY` (preferred) before calling `ElectricalNetwork.solve_load_flow()`. More information in the documentation.
  - Several methods on the `FlexibleParameter` class that previously required `auth` are changed. Make sure to follow the documentation to update your code.
- #163 #158 Fix `ElectricalNetwork.res_transformers` returning an empty dataframe when max_power is not set.
- #163 Several unused exception codes were removed. An `EMPTY_NETWORK` code was added to indicate that a network is being created with no elements.
- #163 Remove the `ElectricalNetwork.res_info` attribute. `ElectricalNetwork.solve_load_flow()` now returns the tuple (number of iterations, residual).
- #163 Remove the `Bus.clear_short_circuits()` and `ElectricalNetwork.clear_short_circuits()`  methods. It is currently not possible to clear short-circuits from the network.
- #163 Improve performance of network creation and results access.
- #163 Attributes `phases` and `bus` are now read-only on all elements.
- #151 Require Python 3.10 or newer.


alihamdan (approved):
Good to have the final release.
I’d also mention that the documentation has moved and link the new location in the release notes.

Saelyos (approved):

PR #189, opened 2024-02-26
Null active power for flexible loads

Saelyos:
Allow flexible loads to have a null active theoretical power.

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -22,6 +22,7 @@ myst:
   Voltage, current, and power limits are added to the buses, lines, and transformers.
   The line parameters IDs are also updated to match the new line parameters catalogue.
 - {gh-pr}`182` Improve the error message when trying to access results on the network before running the load flow.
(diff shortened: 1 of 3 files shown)


roseau/load_flow/models/tests/test_flexible_parameters.py
```diff
@@ -254,63 +262,3 @@ def test_flexible_parameter():
...
-def projection(request) -> Projection:
-    return Projection(type=request.param)
-
-
-@pytest.fixture()
-def flexible_parameter(control_p, control_q, projection) -> FlexibleParameter:
-    return FlexibleParameter(control_p=control_p, control_q=control_q, projection=projection, s_max=Q_(5, "kVA"))
-
-
-def test_plot(flexible_parameter):
```
alihamdan, line 289: Do we still have this test somewhere?

Saelyos: I've moved it to the engine part, as it was using the results

alihamdan (approved):
Thanks.

PR #192, opened 2024-03-04
Optimize results

alihamdan:
This PR includes some optimizations that together added up to 3x performance improvements in a timeseries study I am doing especially on line results. The main things done are:
1. Caching whenever possible for things that we know are immutable like phases, z_line and its inverse matrices, voltage_phases, etc.
2. Avoiding the use of pint `.m_as` as much as possible. When using pint quantities, accessing the value using `.m` is much faster in tight loops when we know that the unit we want is the default one. Also avoiding the whole pint conversion interface (using `ureg_wraps` in our case) contributed the most to the performance improvements as it avoids the double conversion to a quantity then to a value.
3. Inlining results access in the `res_` properties on the electrical network to avoid calling `_res_currents_getter` several times for example.
4. A few other micro-optimizations

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -15,7 +15,7 @@ myst:
 
 ## Unreleased
 
-- {gh-pr}`191` Speed up results access by up to 3x using several optimization techniques. This is especially
+- {gh-pr}`192` Speed up results access by up to 3x using several optimization techniques. This is especially
   noticeable in timeseries simulations and when accessing results of large networks.
 - {gh-pr}`184` Improve the documentation to have a better SEO (sitemap, metadata and canonical URLs). The navigation
   menu has also been improved.
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -15,6 +15,8 @@ myst:
 
 ## Unreleased
 
+- {gh-pr}`191` Speed up results access by up to 3x using several optimization techniques. This is especially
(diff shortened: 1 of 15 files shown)


doc/Changelog.md
```diff
@@ -15,6 +15,8 @@ myst:
 
 ## Unreleased
 
+- {gh-pr}`191` Speed up results access by up to 3x using several optimization techniques. This is especially
```
alihamdan, line 18:
```suggestion
- {gh-pr}`192` Speed up results access by up to 3x using several optimization techniques. This is especially
```

roseau/load_flow/models/tests/test_loads.py
```diff
@@ -204,7 +204,7 @@ def test_flexible_load():
     fp = [fp_pq_prod, fp_const, fp_const]
     with pytest.raises(RoseauLoadFlowException) as e:
         PowerLoad("flexible load", bus, powers=[10 - 250j, 0, 0j], phases="abcn", flexible_params=fp)
-    assert "The reactive power is lesser than the parameter q_min for flexible load" in e.value.msg
+    assert "The reactive power is less than the parameter q_min for flexible load" in e.value.msg
```
benoit9126, line 207: `lower than`?

benoit9126 (approved):

PR #203, opened 2024-03-18
Detect invalid element overrides

### LINKED ISSUE #186: ENH: Detect invalid element overrides
We should detect when multiple elements with same type and ID are being created and raise an error. For example, this code should raise an error:
  ```python
  import roseau.load_flow as lf
  bus = lf.Bus("bus1", phases="an")
  load1 = lf.PowerLoad("load1", bus, powers=[1000])
  load2 = lf.PowerLoad("load1", bus, powers=[1000])  # <- oops, forgot to change the ID
  ```
The same problem could happen when working in notebooks if a cell that creates a load is called multiple times.

Preventing this helps avoid problems that are difficult to debug because the duplicate elements will not appear in the dictionaries of the `ElectricalNetwork` class, but they exist in the internal structure representing the low level network.

**Implementation details**
To do this, we can replace the `_connected_elements` private attribute on the `Element` class by separate dictionaries, one per element type (similar to the electrical network loads, buses, etc dictionaries). We can also add an iterator method `_iter_connected_elements` that yields from these dictionaries for easy access to all elements.

alihamdan:
Fixes #186


Saelyos:
Hello, I'm not sure of the problem you are trying to solve. 

If we look at the example linked in #186, and if we run the following code: 
```python
import roseau.load_flow as lf
bus = lf.Bus("bus1", phases="an")
load1 = lf.PowerLoad("load1", bus, powers=[1000])
load2 = lf.PowerLoad("load1", bus, powers=[1000])  # <- oops, forgot to change the ID
en = lf.ElectricalNetwork.from_element(bus)
```
We get the right error message: 
```
roseau.load_flow.exceptions.RoseauLoadFlowException: Duplicate ID for an load in this network: 'load1'. [bad_load_id]
```

However the problem could arise if we connect the load later on (I had to add a few elements):
```python
import roseau.load_flow as lf
bus = lf.Bus("bus1", phases="an")
vs = lf.VoltageSource("source", bus, voltages=[10])
pref = lf.PotentialRef("pref", bus)
load1 = lf.PowerLoad("load1", bus, powers=[1000])
en = lf.ElectricalNetwork.from_element(bus)
load2 = lf.PowerLoad("load1", bus, powers=[1000])  # <- oops, forgot to change the ID
```
In that case no error message is currently raised, but it should be easy to add in the `ElectricalNetwork._connect_element` method something like:
```python
elif isinstance(element, AbstractLoad):
    if element.id in self.loads:
        raise RoseauLoadFlowException("Duplicate ID...", code=error_code)
    self.loads[element.id] = element
```
This seems simpler than modifying every elements `_connected_elements` attributes. What do you think ?

alihamdan:
> However the problem could arise if we connect the load later on

Yes, this is the main problem we are trying to solve.

> This seems simpler than modifying every elements `_connected_elements` attributes. What do you think ?

You are right, and that's why the PR is still a draft 😉.
I started by changing the `_connected_elements` to a dictionary and adding the check there following the suggeston of Benoît but then realized that this does not detect all cases and we have to modify the network's `_connect_element` method instead anyway.

roseau/load_flow/network.py
```diff
@@ -1121,22 +1121,34 @@ def _connect_element(self, element: Element) -> None:
...
+            container, element_type = self.grounds, "ground"
         elif isinstance(element, PotentialRef):
-            self.potential_refs[element.id] = element
+            container, element_type = self.potential_refs, "potential reference"
         else:
             msg = f"Unknown element {element} can not be added to the network."
             logger.error(msg)
             raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_ELEMENT_OBJECT)
+        if element.id in container and container[element.id] is not element:
+            element._disconnect()  # Don't leave it lingering in other elemnets' _connected_elements
```
benoit9126, line 1145: Small typo: `elemnets`

benoit9126 (approved):

Saelyos (approved):

PR #204, opened 2024-03-18
Remove restrictions on geometry types and allow custom CRS

### LINKED ISSUE #193: ENH: Remove checks on the type of geometry accepted by an element
We currently force the type of geometry accepted by different elements. For example, buses, switches, and transformers geometries have to be Points, lines geometries have to be LineStrings. While this might make sense in the general case, we should leave it up to the user to choose how they represent their elements.

alihamdan:
Resolves #193

- Remove the restriction that the geometry of a bus must be a point, etc
- Add a new `crs` parameter to the network's constructor to allow the user to specify the CRS of their geometries, default set to `"EPSG:4326"`

I also noticed  spurious `**kwargs` in constructors remaining from the SaaS version where we needed to pass arbitrary keywords to elements. I removed them as they are no longer used.

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -15,6 +15,7 @@ myst:
 
 ## Unreleased
 
+- {gh-pr}`204` {gh-issue}`193` Remove restrictions on geometry types. Allow specifying the CRS of the geometries.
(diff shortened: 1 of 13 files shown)


benoit9126 (approved):
:+1:

PR #223, opened 2024-05-29
Fix from_catalogue textual match with literal string inputs

### LINKED ISSUE #222: BUG: line parameters from catalogue fails to match the specified name
To reproduce:
```pycon
>>> import roseau.load_flow as rlf
>>> rlf.LineParameters.from_catalogue("O_AL_12")
RoseauLoadFlowException: Several line parameters matching the query (name='O_AL_12') have been found: 'O_AL_12', 'O_AL_120'. [catalogue_several_found]
```

Expected:
The method to create a `LineParameters` object for the catalogue entry with the same name.

alihamdan:
Fixes #222

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -19,6 +19,9 @@ og:description: See what's new in the latest release of Roseau Load Flow !
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


benoit9126 (approved):
Perfect!

PR #224, opened 2024-06-07
Update the docs

alihamdan:
- Update the docs to use the convention `import roseau.load_flow as rlf`
- Modify the `show_versions` function to show the version of `roseau-load-flow`
- Minor tweaks to the Getting Started page 
- Run ruff formatter on code blocks inside docstrings
- Change the warning on outdated results *of an element* to include the element type and ID instead of the generic "The results of this element may be outdated"

### DIFF
--- a/doc/usage/Connecting_Elements.md
+++ b/doc/usage/Connecting_Elements.md
@@ -22,23 +22,23 @@ Let's take the electrical network of the [Getting started page](usage-getting-st
 
 ```pycon
 >>> import numpy as np
-... from roseau.load_flow import *
+... import roseau.load_flow as rlf
 
->>> source_bus = Bus(id="sb", phases="abcn")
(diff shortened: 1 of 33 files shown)


doc/usage/Connecting_Elements.md
```diff
@@ -84,7 +84,7 @@ The `disconnect` method is only available for loads and for voltage sources.
 >>> load.disconnect()
 ```
 
-Now, the load does not belong anymore to the network `en`. Symmetrically, the network doesn't have this load anymore:
+Now, the loads no longer belongs to the network `en`. Symmetrically, the network doesn't have this load anymore:
```
alihamdan, line 87:
```suggestion
Now, the load no longer belongs to the network `en`. Symmetrically, the network doesn't have this load anymore:
```

benoit9126 (approved):

PR #226, opened 2024-06-11
Add a page about data exchange to the docs

alihamdan:
The page lists currently supported converters to RLF


doc/usage/Data_Exchange.md
```diff
@@ -0,0 +1,141 @@
...
+  html_meta:
+    "description lang=en": |
+      Roseau Load Flow data exchange formats: Save/Load a network (JSON); convert from Power Factory (DGS),
+      OpenDSS, etc.
+    "description lang=fr": |
+      Conversion vers Roseau Load Flow: Enregistrer/Charger un réseau au format JSON, convertir depuis Power
+      Factory (DGS) et OpenDSS.
+    "keywords lang=fr": simulation, réseau, électrique, Power Factory, OpenDSS, JSON
+    "keywords lang=en": simulation, distribution grid, Power Factory, OpenDSS, JSON
+---
```
alihamdan, line 12: @mullerlo could you please take a look at the page metadata here and tell me if you have any suggestions

mullerlo: Looks good to me !

roseau/load_flow/io/dgs.py
```diff
@@ -378,7 +395,7 @@ def _generate_loads(
 
         # Balanced or Unbalanced
         s = [s_phase / 3, s_phase / 3, s_phase / 3] if sa == 0 and sb == 0 and sc == 0 else [sa, sb, sc]
-        loads[load_id] = PowerLoad(id=load_id, phases="abcn", bus=buses[bus_id], powers=s)
+        loads[load_id] = PowerLoad(id=load_id, bus=buses[bus_id], powers=s)
```
alihamdan, line 398: @Saelyos this change and the change above for the source are made because we don't allow floating neutrals yet and the DGS import fails if the bus has "abc" phases while the load is defined as "abcn". Is this correct or should we always keep "abcn" here and allow floating neutrals?

benoit9126 (approved):
Looks good to me for now. Two remarks:
- You could track the pfd files with git LFS: `git lfs track '*.pfd'`
- We need to add the “floating neutral” feature to the Python part of RLF. Maybe not now, but the sooner, the better. When can we discuss of this feature with @Saelyos ?

roseau/load_flow/io/dgs.py
Saelyos: For the voltage source, we can keep the same phases as the bus, but the voltages need to be adapted for the delta voltage source. 

For the load, I'm actually not sure what the loads are supposed to be in the dgs file. We have assumed for now that the loads are wye loads, but I'm not sure it's true : there is a `phtech` attribute that we don't take into account.

PR #237, opened 2024-07-09
Sphinx configuration improvements

benoit9126:
Small improvements of the Sphinx configuration:
- Delete the `autodoc` configuration part (we are using `autoapi`)
- Add some Intersphinx mappings
- Delete the warning for the `roseau.load_flow_engine.cy_engine` module which is not found when generating the documentation.

@alihamdan By the way, the command `make linkcheck` already exists and checks all the external links found in the documentation. We should run it sometimes to ensure the documentation is up-to-date.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     hooks:
       - id: poetry-check
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.5.1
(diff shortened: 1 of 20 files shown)


benoit9126:
There is still a bug in `sphinx-autoapi` readthedocs/sphinx-autoapi#473 that needs to be solved to have a better documentation (ie. API doc similar to the "API Reference" section for each model)

PR #239, opened 2024-07-11
Load type in the JSON format (version 2)

### LINKED ISSUE #238: ENH: Add the load type to the json file
The load type is not added to the JSON file. Currently we check for the existence of a `powers` key in the load dictionary to know if the load is a `PowerLoad`, `currents` key for `CurrentLoad`, etc. It would be nicer for the GUI if we had this field directly in the file. Additionally, it will be also useful to add the same field to the results JSON file as in this file it is impossible to know the type of the load used.

This can be done as part of version 2 transition.

benoit9126:
Close #238


roseau/load_flow/models/loads/loads.py
```diff
@@ -221,11 +222,11 @@ def from_dict(cls, data: JsonDict, *, include_results: bool = True) -> "Abstract
...
+        elif load_type == "current":
+            currents = [complex(i[0], i[1]) for i in data["currents"]]
             self = CurrentLoad(id=data["id"], bus=data["bus"], currents=currents, phases=data["phases"])
-        elif (z_list := data.get("impedances")) is not None:
-            impedances = [complex(z[0], z[1]) for z in z_list]
+        elif load_type == "impedance":
+            impedances = [complex(z[0], z[1]) for z in data["impedances"]]
             self = ImpedanceLoad(id=data["id"], bus=data["bus"], impedances=impedances, phases=data["phases"])
         else:
             msg = f"Unknown load type for load {data['id']!r}"
```
alihamdan, line 232: Since we have the type now
```suggestion
            msg = f"Unknown load type {load_type!r} for load {data['id']!r}"
```

alihamdan (approved):
Thanks. Just one small nit.

PR #247, opened 2024-07-31
Add `connect_neutral` parameter to loads and sources

alihamdan:
This gives the users the ability to create loads with connected neutral and loads with floating neutral on the same bus without needing to create a dedicated "abc" bus just to have a floating neutral connection.

The default behavior of automatically considering the neutral as floating when the bus does not have a neutral is conserved.


roseau/load_flow/models/loads/loads.py
```diff
@@ -53,21 +53,32 @@ def __init__(self, id: Id, bus: Bus, *, phases: str | None = None) -> None:
...
+
+            connect_neutral:
+                Specifies whether the load's neutral should be connected to the bus's neutral or
+                left floating. By default, the load's neutral is connected when the load has a
+                neutral. If the bus does not have a neutral, the load's neutral is left floating
+                by default. To override the default behavior, pass an explicit ``True`` or ``False``.
         """
         if type(self) is AbstractLoad:
             raise TypeError("Can't instantiate abstract class AbstractLoad")
         super().__init__(id)
```
Saelyos, line 68: Maybe an error should be raised for some cases, for example if `connect_neutral` is True, but there are no neural in the bus or in the phases.

roseau/load_flow/models/loads/loads.py
```diff
@@ -53,21 +53,32 @@ def __init__(self, id: Id, bus: Bus, *, phases: str | None = None) -> None:
...
             phases = bus.phases
+            if connect_neutral is None:
+                connect_neutral = True
         else:
             self._check_phases(id=id, phases=phases)
             # Also check they are in the bus phases
             phases_not_in_bus = set(phases) - set(bus.phases)
+            if connect_neutral is None:
+                connect_neutral = "n" in bus.phases or "n" not in phases
             # "n" is allowed to be absent from the bus only if the load has more than 2 phases
```
Saelyos, line 79: Shouldn't this be `connect_neutral = "n" in bus.phases and "n" in phases` ?

Saelyos (changes requested):
While I understand the goal of `connect_neutral` for loads, I'm not sure of the use case for the voltage sources, why would you want to connect only `abc` but not the neutral ?

Saelyos (approved):
Thanks for the modifications, it looks good for me now

PR #255, opened 2024-08-06
Update the figures of the documentation

benoit9126:
I updated the LaTeX figures. The main improvement is for the figures of loads and voltages sources. The position of the `B` phase and the `C` phases have been inverted to be compliant with #249


alihamdan (approved):
Thanks

PR #256, opened 2024-08-07
Allow passing scalar values to loads

### LINKED ISSUE #250: ENH: Allow passing scalar values for load inputs
Similar to #248 but for `PowerLoad` powers, `CurrentLoad` currents, and `ImpedanceLoad` impedances.

For `PowerLoad`, we should write good documentation to avoid confusion on what value of power is expected (for example single-phase vs three-phase).

alihamdan:
Closes #250

### DIFF
--- a/roseau/load_flow/typing.py
+++ b/roseau/load_flow/typing.py
@@ -80,6 +80,7 @@ ComplexArrayLike1D: TypeAlias = (
     | Q_[Sequence[complex | float]]
     | Sequence[complex | float | Q_[complex | float]]
 )
+ComplexArrayLikeScalarOr1D: TypeAlias = ComplexArrayLike1D | Q_[complex | float] | complex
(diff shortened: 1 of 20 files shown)


roseau/load_flow/typing.py
```diff
@@ -80,6 +80,7 @@
     | Q_[Sequence[complex | float]]
     | Sequence[complex | float | Q_[complex | float]]
 )
+ComplexArrayLikeScalarOr1D: TypeAlias = ComplexArrayLike1D | Q_[complex | float] | complex
```
benoit9126, line 83: ` | float` at the end of this type?

benoit9126 (approved):

roseau/load_flow/typing.py
alihamdan: `complex` is equivalent to `complex | float | int`

PR #266, opened 2024-09-04
DOC: Update the tutorials

benoit9126:
I updated the tutorials using the last version of RLF. I also add some README files


alihamdan:
This is blocked on the current changes that break the API, converted to draft.

alihamdan:
I updated the tutorials repository. I suggest that we remove the tutorials from this repository and link to the other repository instead when we make it public after the next release.

I know I advocated for including the tutorials here before but there are things that I didn't consider before that would affect beginners, namely:
- Problems with git lfs on Windows as this repository uses lfs because it includes large pdf files. I only noticed these problems recently when working on another project on Windows with Louise
- Downloading the tutorials as a zip file as an alternative for people who don't have git installed (mainly Windows users :/). This is what most users comming from https://github.com/Team-Nando/Tutorial-DERHostingCapacity-0-dss_python?tab=readme-ov-file#run-part-0 will be doing
- Installing the wrong dependencies by running `uv sync` in the root here instead of the tutorials folder

I'll send a PR to remove the tutorials from here if there are no objections to the proposal @benoit9126.

PR #274, opened 2024-10-23
Allow a ground to be connected to multiple phases of the same bus with a warning

alihamdan:
This used to raise before because it is almost certainly an error. We needed it in some of the niche cases we were testing so now it warns instead.


alihamdan:
Never mind, we can achieve the same thing with a short circuit on the bus.

PR #280, opened 2024-11-13
Modify the voltage limits of buses

### LINKED ISSUE #278: ENH: Nominal voltage of buses
**Describe the feature you want. What problem does it solve?**

I would like to store in the Bus object the nominal voltage of the bus. It has several consequences:
* a `res_voltage_level` property can be added to the bus (return `np.nan` if no `nominal voltage`)
* The `min_voltage` and the `max_voltage` could be replaced with percentages of the nominal voltage (with a warning if it is too large, such as a voltage)
* If a percentage is given to `min_voltage` and the `max_voltage`, `nominal_voltage` becomes mandatory.
* The `res_buses_voltages` data frame could be filled with an additional column `voltage_level`
* the `buses` data frame could be filled with an additional column `nominal_voltage`
* The method `propagate_limits` could propagate the limits and the nominal voltage to the galvanically isolated component (not only the limits)
* The method `propagate_limits` could be renamed `propagate_voltages`

benoit9126:
Close #278

The `Bus` constructor now accepts a nominal voltage (phase-to-phase, in V), a `min_voltage_level` (without unit) and a `max_voltage_level` (without unit). These arguments replace `min_voltage` and `max_voltage`. The `Bus` class now have some additional properties:
- `res_voltages_level`: it retrieves the voltage level of the bus according to the provided phase-to-phase nominal
    voltage;
- `min_voltage_level` and `max_voltage_level`: these properties retrieve the minimum and maximum voltage levels.
- `min_voltage` and `max_voltage`: these properties (that existed before) now compute the minimum and maximum
    voltage (in V) from the nominal voltage and the minimum and maximum levels.

The file format also changed to take into account these changes. If a `min_voltage` or `max_voltage` existed in a file of a previous version, they are lost when upgrading the file.


roseau/load_flow/models/buses.py
```diff
@@ -151,6 +162,22 @@ def res_voltages(self) -> Q_[ComplexArray]:
         """
         return self._res_voltages_getter(warning=True)
 
+    @property
+    def res_voltages_level(self) -> FloatArrayLike1D | None:
```
alihamdan, line 166: The name should be `res_voltage_levels`. The type must be concrete as it is a return type `Q_[NDArray[np.float64]] | None`.

alihamdan (changes requested):
Excellent. I just have a few suggestions.

doc/usage/Getting_Started.md
```diff
@@ -82,16 +83,28 @@ line and a constant power load. This network is a low voltage network (three-pha
...
 ... import roseau.load_flow as rlf
 
->>> # Nominal phase-to-neutral voltage (typical european voltage)
-... un = 400 / np.sqrt(3)  # In Volts
+>>> # Nominal phase-to-phase voltage (typical european voltage)
+... un = 400  # In Volts
 ... # Optional network limits (for analyzing network violations)
-... u_min = 0.9 * un  # V
-... u_max = 1.1 * un  # V
+... u_min_level = 0.9  # V
```
alihamdan, line 89: Same for the max level
```suggestion
... u_min_level = 0.9  # 90% Un
```

alihamdan (approved):
LGTM. One more comment update in the docs.

Saelyos (approved):

PR #283, opened 2024-11-15
`max_currents` and `sections` become arrays

benoit9126:
Following a discussion with @alihamdan, the `max_current` and `section` parameters of `LineParameters` becomes arrays to allow a definition of the section and of the maximum current per conductor. They are also renamed `max_currents` and `sections`. 

Currently, the catalogue of lines is not modified. I propose to modify it when some wires with a half-section neutral will be added. Moreover, the import from PwF and the export to PwF could use a neutral section (if available in PwF).


roseau/load_flow/models/lines/parameters.py
```diff
@@ -81,11 +88,16 @@ def __init__(
...
-                The maximum current loading of the line (A). The maximum current is optional, it is
+            max_currents:
+                The maximum currents loading of the line (A). The maximum currents are optional, they are
                 not used in the load flow but can be used to check for overloading.
                 See also :meth:`Line.res_violated <roseau.load_flow.Line.res_violated>`.
 
+                When a scalar value is provided, it creates an array with the same maximum current for each conductor.
+                The scalar value passed is assumed to be the maximum current of each conductor, not the total
+                multi-phase maximum current. To create a different maximum current per conductor, provide a
+                vector of current values with the same length as the number of conductor of the line.
```
alihamdan, line 99: The "total multi-phase maximum current" doesn't make sense (unlike powers of loads)

```suggestion
                To create a different maximum current per conductor, provide a vector of current values with
                the same length as the number of conductor of the line.
```

roseau/load_flow/models/lines/parameters.py
```diff
@@ -816,7 +919,7 @@ def from_power_factory(  # noqa: C901
         cohl: Literal[0, "Cable", 1, "OHL"] = "Cable",
         conductor: Literal["Al", "Cu", "Ad", "As", "Ds"] | None = None,
         insulation: Literal[0, "PVC", 1, "XLPE", 2, "Mineral", 3, "Paper", 4, "EPR"] | None = None,
-        section: float | Q_[float] | None = None,
+        section: FloatScalarOrArrayLike1D | None = None,
```
alihamdan, line 922: Power factory has a scalar section, this should not change. 
```suggestion
        section: float | Q_[float] | None = None,
```

roseau/load_flow/models/transformers/parameters.py
```diff
@@ -416,7 +416,7 @@ def from_power_factory(
                 PwF parameter `maxload` (Max. Thermal Loading Limit). The maximum loading of the
                 transformer in (%) of the nominal power. This parameter is defined on the transformer
                 element (`ElmTr2`) in PwF instead of the transformer type (`TypTr2`).
-                This is used to compute `max_current` and is used for violation checks.
+                This is used to compute `max_currents` and is used for violation checks.
```
alihamdan, line 419: This is an old typo
```suggestion
                This is used to compute `max_power` and is used for violation checks.
```

roseau/load_flow/models/transformers/parameters.py
```diff
@@ -558,7 +558,7 @@ def from_open_dss(
             normhkva:
                 OpenDSS parameter: `NormHKVA`. Normal maximum kVA rating for H winding (1). Usually
                 100 - 110% of maximum nameplate rating.
-                This value is passed to `max_current` and used for violation checks.
+                This value is passed to `max_currents` and used for violation checks.
```
alihamdan, line 561: Same typo
```suggestion
                This value is passed to `max_power` and used for violation checks.
```

roseau/load_flow/models/lines/parameters.py
```diff
@@ -186,14 +233,68 @@ def insulator_type(self) -> InsulatorType | None:
...
+                values = None
+            elif value < 0:
+                msg = f"Sections must be non-negative: {value} mm² was provided."
+                logger.error(msg)
+                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_SECTIONS_VALUE)
+            else:
+                values = np.array([value for _ in range(self._size)], dtype=np.float64)
+        else:
+            if np.any(value_isna):
+                msg = f"Sections can't be null: [{', '.join(f'{x}' for x in value)}] mm² was provided."
```
alihamdan, line 283: Actually sections can be null (`None`) but if an array-like is given it cannot contain null values.

Nit: you don't need a list comprehension, you can use `tolist()` that returns python floats that are printed  as normal float literals instead of `np.float64(whatever)` (here and everywhere you use this)
```suggestion
                msg = f"Sections cannot contain null values: {value.tolist()} mm² was provided."
```

roseau/load_flow/models/lines/parameters.py
```diff
@@ -186,14 +233,68 @@ def insulator_type(self) -> InsulatorType | None:
...
+                logger.error(msg)
+                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_SECTIONS_VALUE)
+            if len(value) != self._size:
+                msg = f"Incorrect number of sections: {len(value)} instead of {self._size}"
+                logger.error(msg)
+                raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_SECTIONS_SIZE)
+
+            values = np.array(value, dtype=np.float64)
```
alihamdan, line 291: Move this line to the beginning of the `else` block so that we are sure we are manipulating numpy arrays when doing the checks (like when calling `len(value)` above or when calling `value.tolist()` as suggested in the comment above.

roseau/load_flow/models/lines/parameters.py
```diff
@@ -103,6 +115,15 @@ def __init__(
                 it is informative only and is not used in the load flow. This field gets
                 automatically filled when the line parameters are created from a geometric model or
                 from the catalogue.
+
+            sections:
+                The sections of the conductor. The sections are optional, it is informative only and is not used in
```
alihamdan, line 120:
```suggestion
                The sections of the conductor. The sections are optional, they are informative only and are not used in
```

alihamdan (commented):
Looks good. I left a few suggestions about documentation improvement, error messages, and a code nit.

roseau/load_flow/utils/types.py
```diff
@@ -170,19 +169,31 @@ class InsulatorType(StrEnum):
...
+@deprecated("Use the class `Insulator` instead.")
+def InsulatorType(*args, **kwargs):  # noqa: N802
+    """A deprecated class: please use the class `Insulator` instead."""
+    return Insulator(*args, **kwargs)
+
+
+@deprecated("Use the class `Material` instead.")
+def ConductorType(*args, **kwargs):  # noqa: N802
+    """A deprecated class: please use the class `Material` instead."""
+    return Material(*args, **kwargs)
```
alihamdan, line 199: These do not work because now `ConductorType.CU` will raise an exception. I think we can clean this file and add the following to the end of `roseau/load_flow/__init__.py`

```python
def __getattr__(name: str) -> Any:
    deprecated_classes = {"ConductorType": Material, "InsulatorType": Insulator}

    if name in deprecated_classes:
        import warnings

        from roseau.load_flow.utils._exceptions import find_stack_level

        new_class = deprecated_classes[name]
        warnings.warn(
            f"The {name} class is deprecated. Use {new_class.__name__} instead.",
            stacklevel=find_stack_level(),
        )
        return new_class
    else:
        # raise AttributeError with original error message
        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")

```

roseau/load_flow/models/lines/parameters.py
```diff
@@ -81,28 +89,42 @@ def __init__(
...
-                The type of the cable insulator (PVC, XLPE, ...). The insulator type is optional,
-                it is informative only and is not used in the load flow. This field gets
+            insulators:
+                The types of the cable insulator (PVC, XLPE, ...). The insulators are optional,
+                they are informative only and are not used in the load flow. This field gets
                 automatically filled when the line parameters are created from a geometric model or
                 from the catalogue.
+
+            sections:
+                The sections of the conductor. The sections are optional, thy are informative only and ares not used in
```
alihamdan, line 121:
```suggestion
                The sections of the conductor. The sections are optional, thy are informative only and are not used in
```

roseau/load_flow/models/lines/parameters.py
```diff
@@ -398,13 +555,13 @@ def from_geometry(
         id: Id,
         *,
         line_type: LineType,
-        conductor_type: ConductorType | None = None,
-        insulator_type: InsulatorType | None = None,
+        material: Material | None = None,
+        insulator: Insulator | None = None,
```
alihamdan, line 559: Use the plural names to avoid a future breakage when/if we have lines with different material/insulator in the catalogue.

Alternativaly, we can have `material`/`material_neutral` and `insulator`/`insulator_neutral` similar to `section`/`section_neutral`. If you choose this option, we can add these parameters later; no change is needed now.
```suggestion
        materials: Material | None = None,
        insulators: Insulator | None = None,
```

roseau/load_flow/exceptions.py
```diff
@@ -35,12 +35,19 @@ class RoseauLoadFlowExceptionCode(StrEnum):
     BAD_Y_SHUNT_SHAPE = auto()
     BAD_LINE_MODEL = auto()
     BAD_LINE_TYPE = auto()
-    BAD_CONDUCTOR_TYPE = auto()
+    BAD_MATERIAL = auto()
     BAD_INSULATOR_TYPE = auto()
```
alihamdan, line 39: Shouldn't this change as well?
```suggestion
    BAD_INSULATOR = auto()
```

alihamdan (commented):
A few comments

roseau/load_flow/models/lines/parameters.py
benoit9126: I have added all the `neutral` elements in the catalogue (data, `get_catalogue` and `from_catalogue`) and in the `from_geometry` class method.

roseau/load_flow/models/lines/parameters.py
```diff
@@ -677,50 +708,27 @@ def _from_geometry(
             material = _DEFAULT_MATERIAL[line_type]
         if insulator is None:
             insulator = _DEFAULT_INSULATOR[line_type]
+        if material_neutral is None:
+            material_neutral = material
+        if insulator_neutral is None:
+            insulator_neutral = insulator
```
alihamdan, line 714: So here is the problem with this. `insulator_neutral = None` is confusing because we cannot know if it means "no insulation" or it means "no value provided, please use the value of the phases insulator". In other words, the user cannot create a non-insulated neutral with insulated phases.

I suggest we keep the `NONE` enumeration on the `Insulator` class to avoid this confusion here or elsewhere.

roseau/load_flow/__init__.py
```diff
@@ -112,3 +113,28 @@
...
+
+        new_class = deprecated_classes[name]
+        if name not in __rlf_deprecated_classes:
+            import warnings
+
+            warnings.warn(
+                f"The {name} class is deprecated. Use {new_class.__name__} instead.",
+                stacklevel=find_stack_level(),
+            )
+        __rlf_deprecated_classes.add(name)
```
alihamdan, line 135: Simpler implementation
```suggestion

def __getattr__(name: str) -> Any:
    deprecated_classes = {"ConductorType": Material, "InsulatorType": Insulator}

    if name in deprecated_classes:
        import warnings

        from roseau.load_flow.utils._exceptions import find_stack_level

        new_class = deprecated_classes[name]
        warnings.warn(
            f"The {name} class is deprecated. Use {new_class.__name__} instead.",
            stacklevel=find_stack_level(),
        )
        globals()[name] = new_class
```

alihamdan (approved):

PR #304, opened 2024-12-31
Move important modules to top-level

alihamdan:
This PR re-organizes the `utils` subpackage. The important modules are moved to top-level while internal modules are kept in `utils`.

This is to separate modules that are meant for user use from modules that are meant for internal use. This way we are able to exclude the internal modules from the documentation and make using the important modules easier at the same time.

Old modules are deprecated and will be removed in a future version.


.github/workflows/ci.yml
```diff
@@ -17,6 +17,7 @@ on:
 env:
   CI: true
   UV_SYSTEM_PYTHON: 1
+  COVERAGE_CORE: sysmon
```
alihamdan, line 20: Tested locally on Python 3.13 (this works on 3.12+). The time of execution of:
- `pytest`: ~5 seconds
- `pytest --cov` without this change: ~13 seconds
- `pytest --cov` with this change: ~5 seconds

Basically, this removes the overhead of coverage. It requires setting `branch = false` in `[tool.coverage.run]`.

PR #310, opened 2025-01-06
Support star and zigzag transformers with non brought out neutrals

### LINKED ISSUE #308: BUG: The neutral of `Dy11` transformer should not be connected
A `Dyn11` transformer has a brought out neutral while `Dy11` does not. The neutral of the latter must not be connected to the bus.

alihamdan:
Fixes #308


benoit9126 (approved):

roseau/load_flow/models/transformers/transformers.py
```diff
@@ -209,8 +211,11 @@ def _compute_phases_three(
...
         phases2: str | None,
     ) -> tuple[str, str]:
-        w1_has_neutral = "y" in parameters.winding1.lower() or "z" in parameters.winding1.lower()
-        w2_has_neutral = "y" in parameters.winding2.lower() or "z" in parameters.winding2.lower()
+        w1 = parameters.winding1
+        w2 = parameters.winding2
+        clock = parameters.phase_displacement
+
+        w1_has_neutral = w1.endswith("N")
         if phases1 is None:
```
Saelyos, line 219: Should we emit a warning if `n` is in `bus1.phases` but `w1_has_neutral` is False ?

PR #325, opened 2025-01-30
Add branch potentials results

alihamdan:
This PR makes a lot of internal changes but has no visible effect on the users.

The models classes have been refactored:
1. Added a `BaseTerminal` class that represents single-port elements (e.g buses and loads/sources). A terminal has `phases`, `res_potentials`, and `res_voltages`. `Bus` inherits from this class
2.  Added a `BaseConnectable` class that represents connectable/disconnectable elements (e.g loads and sources). This class inherits from `BaseTerminal` and additionally has `bus`, `disconnect()`, floating neutral logic, `res_currents`, and `res_powers`. `AbstractLoad` and `VoltageSource` inherit from this class
3. Modified the `AbstractBranch` objects to fetch potentials from their cython objects instead of the buses similar to what was done to loads/sources

This is in preparation for adding the ability to connect the ground to other elements than a bus. The refactoring will be especially useful with floating neutrals for transformers, impedant grounds, and other upcoming chnages.


benoit9126 (approved):
Major changes, greatly simplified! :+1:

Saelyos (approved):
Looks good (I like the addition of `Generic[_CyE]`)
Just two small comments for the readability of the code

PR #326, opened 2025-01-31
Rename `Bus.potentials` to `Bus.initial_potentials`

alihamdan:
Similar to `Bus.initial_voltage` in rlfs.

This is to avoid confusion since the notion of fixed bus voltage/potentials exist for slack buses, which is not is intended for this parameter.


alihamdan:
Open question: should we rename the field in the JSON file?

benoit9126 (approved):

benoit9126:
I think yes. To keep consistency.

Saelyos (approved):

PR #327, opened 2025-02-03
Support ground connections for loads, sources and transformers

alihamdan:


Saelyos:
I'm wondering about the global design of ground connections, as the changes in this PR do not address connections with impedances easily. 

Could a `Connection` object be a good design ? It would have an internal C++ object, and it also allows to have disconnection and results getter. 

Something like 
```python
con = Connection(bus, ground)   # or con = Connection(bus, ground, z=01+01j)

# network creation, solving...

current = con.res_current  # Current flowing through the ground connection

con.disconnect()  # Same as load.disconnect(), the connection isn't used anymore
```

PR #335, opened 2025-02-13
Add support for impedant ground connections and grounding for terminals and branches

alihamdan:
Add a `GroundConnection` element to create potentially impedant connections to the ground.

Supersedes and closes #327


benoit9126 (approved):

roseau/load_flow/models/grounds.py
```diff
@@ -127,3 +129,248 @@ def _to_dict(self, include_results: bool) -> JsonDict:
...
+        if np.isclose(self._impedance, 0):
+            if not isinstance(self._cy_element, CySwitch):
+                if self._cy_element is not None:
+                    self._cy_element.disconnect()
+                self._cy_element = CySwitch(n=1)
+                self._cy_connect()
+            else:
+                pass  # do nothing
+        else:
+            z_line = np.array([[self._impedance]], dtype=np.complex128)
```
Saelyos, line 295: Somehow it works most of the time, but `z_line` should be a an array of only one dimension when passed to Cython (I've had some problems locally with `Buffer has wrong number of dimensions (expected 1, got 2)`)

roseau/load_flow/models/grounds.py
```diff
@@ -127,3 +129,248 @@ def _to_dict(self, include_results: bool) -> JsonDict:
...
+    @ureg_wraps(None, (None, "ohm"))
+    def impedance(self, value: Complex | Q_[Complex]) -> None:
+        self._impedance = complex(value)
+        self._invalidate_network_results()
+        if np.isclose(self._impedance, 0):
+            if not isinstance(self._cy_element, CySwitch):
+                if self._cy_element is not None:
+                    self._cy_element.disconnect()
+                self._cy_element = CySwitch(n=1)
+                self._cy_connect()
```
Saelyos, line 291: As we change the `CyElement` here, the `CyElectricalNetwork` of the network isn't valid anymore. We have to mark the network as invalid (as we do when we add a short circuit for example) with 
```python
if self.network is not None:
    network._valid = False
```

roseau/load_flow/models/grounds.py
```diff
@@ -127,3 +129,248 @@ def _to_dict(self, include_results: bool) -> JsonDict:
...
+            else:
+                pass  # do nothing
+        else:
+            z_line = np.array([[self._impedance]], dtype=np.complex128)
+            if not isinstance(self._cy_element, CySimplifiedLine):
+                if self._cy_element is not None:
+                    self._cy_element.disconnect()
+                self._cy_element = CySimplifiedLine(n=1, z_line=z_line)
+                self._cy_connect()
```
Saelyos, line 300: Same here, the network is now invalid.

Saelyos (changes requested):
The interface is good for me. 
There is just a problem of `CyElectricalNetwork` re-creation, this is an exemple of a test that result in a segfault currently :
```python
bus = Bus("bus", phases="abcn")
VoltageSource("vs", bus, voltages=20e3)
ground = Ground("ground")
PotentialRef("pref", ground)
gc = GroundConnection(ground=ground, element=bus)

en = ElectricalNetwork.from_element(bus)
en.solve_load_flow()

# Non-zero impedance
gc.impedance = 0.5 + 0.1j
en.solve_load_flow()
```

Saelyos (approved):

Issue #323, opened 2025-01-08
[single] Add short-circuit support


alihamdan:
Short-circuits are not supported yet in RLFS. We can add support for symmetrical short-circuits like three-phase and three-phase-to-ground. Asymmetrical short-circuits will only be supported in RLF.

PR #339, opened 2025-02-21
Add single short circuits

Saelyos:
- Add single short circuit #323 
- Fix error message of elements not passed to the constructor

### DIFF
--- a/roseau/load_flow/models/buses.py
+++ b/roseau/load_flow/models/buses.py
@@ -315,9 +315,8 @@ class Bus(AbstractTerminal[CyBus]):
             raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_PHASE)
         for element in self._connected_elements:
             if isinstance(element, (PowerLoad, CurrentLoad)):
-                load_type = "power" if isinstance(element, PowerLoad) else "current"
                 msg = (
-                    f"A {load_type} load {element.id!r} is already connected on bus {self.id!r}. "
+                    f"A {element.type} load {element.id!r} is already connected on bus {self.id!r}. "
                     f"It makes the short-circuit calculation impossible."
                 )
                 logger.error(msg)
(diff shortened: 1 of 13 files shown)


roseau/load_flow_single/models/buses.py
```diff
@@ -208,6 +210,30 @@ def max_voltage(self) -> Q_[float] | None:
             else Q_(self._max_voltage_level * self._nominal_voltage, "V")
         )
 
+    @property
+    def short_circuit(self) -> bool:
+        """Return the list of short-circuits of this bus."""
```
alihamdan, line 215: The docstring needs to be changed

roseau/load_flow_single/models/buses.py
```diff
@@ -208,6 +210,30 @@ def max_voltage(self) -> Q_[float] | None:
...
+        return self._short_circuit
+
+    def add_short_circuit(self) -> None:
+        """Add a short-circuit by connecting all the phases together with a ground."""
+        from roseau.load_flow_single import PowerLoad
+
+        for element in self._connected_elements:
+            if isinstance(element, PowerLoad):
```
alihamdan, line 223: Last time we discussed this we said that current loads are also not allowed here right?

alihamdan (changes requested):
Thanks, just needs a docstring update and handling current loads. Current loads may need handling in the multiphase version as well.

roseau/load_flow_single/models/buses.py
Saelyos: Oh yes I've forgotten that, it's because we multiply by `u / std::abs(u)` which is `nan` in case of short circuit

roseau/load_flow/models/buses.py
```diff
@@ -314,9 +314,10 @@ def add_short_circuit(self, *phases: str, ground: "Ground | None" = None) -> Non
             logger.error(msg)
             raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_PHASE)
         for element in self._connected_elements:
-            if isinstance(element, PowerLoad):
+            if isinstance(element, (PowerLoad, CurrentLoad)):
+                load_type = "power" if isinstance(element, PowerLoad) else "current"
```
alihamdan, line 318: You could use `element.type`

Issue #334, opened 2025-02-13
ENH: Create a `network_to_dgs` function


benoit9126:
It would be nice to have a `network_to_dgs` function available in order to export a RLF network to the DGS format

alihamdan:
This is implemented for rlfs in #342 

For rlf, the following changes will also be needed (listing here so that we don't forget them):
1. Add an `encoding` parameter to the methods that read/write dgs files
2. Rename the dgs methods similar to rlfs
3. Fix the transformer tap setting and update our DGS test files with the correct `nntap`
4. Read the angle of the voltage source voltage similar to rlfs
5. Investigate more the phases of elements. Phase information is stored on the cubicle but is read-only. Some loads have read-only phase information while others can change their phase technology
6. See if we can factorize more the DGS code between rlf and rlfs

PR #342, opened 2025-03-04
Add DGS import/export to load_flow_single

alihamdan:
Towards #334

This PR adds import/export to DGS format to the single-phase library. Export files have been successfully loaded into PwF and the load flow succeeded. A test case is written to ensure roundtrip conversion to dgs.  More unit test could be added later to cover edge cases.

There are a couple of changes that need to be made in the multi-phase library but they will have to wait for now.

### DIFF
--- a/roseau/load_flow_single/models/core.py
+++ b/roseau/load_flow_single/models/core.py
@@ -4,7 +4,6 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, ClassVar, Generic, NoReturn
 
 import shapely
-from shapely.geometry import shape
 from shapely.geometry.base import BaseGeometry
 from typing_extensions import TypeVar
 
@@ -48,6 +47,7 @@ class Element(ABC, Identifiable, JsonMixin, Generic[_CyE_co]):
         self._fetch_results = False
         self._no_results = True
         self._results_valid = True
+        self._element_info = f"{cls.element_type} {id!r}"  # for logging
 
     @property
     def network(self) -> "ElectricalNetwork | None":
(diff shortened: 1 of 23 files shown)


roseau/load_flow_single/models/core.py
```diff
@@ -167,14 +167,30 @@ def _res_getter(self, value: _T | None, warning: bool) -> _T:
...
+        # In practice, there are no shapely.Geometry objects that are not BaseGeometry objects.
+        if geometry is None:
+            return None
+        if not isinstance(geometry, BaseGeometry):
+            msg = (
+                f"The geometry of {self._element_info} is not a valid shapely geometry. Got {type(geometry).__name__}."
+            )
+            logger.error(msg)
+            raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.BAD_GEOMETRY_TYPE)
+        return geometry
```
Saelyos, line 184: I'm not convinced of this code, if we accept only `BaseGeometry` and not `Geometry` only, shouldn't we use the first to make things clear ? 
I've searched on github and most people seem to use `BaseGeometry` anyway:
`/(?-i)from shapely import.* Geometry[,\n]/ ` -> 138 results
`/(?-i)from shapely.geometry.base import BaseGeometry[,\n]/` -> 2.3k results

Saelyos (approved):

roseau/load_flow_single/network.py
```diff
@@ -1021,3 +1024,75 @@ def _results_to_dict(self, warning: bool, full: bool) -> JsonDict:
...
+            use_name_as_id:
+                If True, use the name of the elements (the ``loc_name`` field) as their id. Otherwise,
+                use the id from the DGS file (the ``FID`` field). Only use if you are sure the names
+                are unique. Default is False.
+
+        Returns:
+            The constructed network.
+        """
+        with open(path, encoding="ISO-8859-10") as f:
+            data = json.load(f)
```
benoit9126, line 1077: For the encoding, I think that hard-coding `"ISO-8859-10"` is tough… 

In the `to_dgs_file` method below, you don't use the encoding keyword argument of the [built-in `open` function](https://docs.python.org/3/library/functions.html#open). It means that on Linux, for instance, when writing a DGS file, `open` will use the default system encoding (`"UTF-8"`) while reading will use the hard-coded `"ISO-8859-10"`...

I propose to add an `encoding` keyword to the class method `from_dgs_file` which is `None` by default. It can be transferred to the `open` function. 

To improve this, we can use a third-party library to detect it (such as [charset_normalizer](https://github.com/jawah/charset_normalizer)), if not explicitly provided by the user.


For the `to_dgs_file` method, I also propose to add an `encoding` parameter to pass to the `open` function. (`None` by default)

roseau/load_flow_single/models/core.py
benoit9126: It seems that Base Geometry is a subclass of Geometry. See [here](https://github.com/shapely/shapely/blob/6166fd54bfc72d5be2394dfc03bbcddb6016a62a/shapely/geometry/base.py#L97). As far as I remember, it has not always been the case...

Anyway, we could be able to use `isinstance(geometry, Geometry)` everywhere. `Geometry` has the advantage to be exposed in the shapely module directly.

alihamdan: I changed the comment and restored the `BaseGeometry` annotations.

benoit9126 (approved):

PR #348, opened 2025-03-07
Add inner currents and powers results for the loads

Saelyos:
The loads have two new properties: `res_inner_currents` and `res_inner_powers` to get the currents and powers that flow in the inner components of the loads.


alihamdan (approved):
Great, thanks.

I just have one suggestion about the documentation.
We also need tests in engine if not already done.

roseau/load_flow/models/loads/loads.py
```diff
@@ -66,12 +66,39 @@ def __init__(self, id: Id, bus: Bus, *, phases: str | None = None, connect_neutr
...
+    @property
+    @ureg_wraps("A", (None,))
+    def res_inner_currents(self) -> Q_[ComplexArray]:
+        """The load flow result of the currents that flow in the inner components of the load (A)."""
+        return self._res_inner_currents_getter(warning=True)
+
+    @property
+    @ureg_wraps("VA", (None,))
+    def res_inner_powers(self) -> Q_[ComplexArray]:
+        """The load flow result of the powers that flow in the inner components of the load (VA)."""
```
alihamdan, line 99:
```suggestion
        """The load flow result of the powers that flow in the inner components of the load (VA).
        
        Unlike `res_powers`, the inner powers do not depend on the reference of potentials. They
        are the physical powers consumed by each of the load dipoles.
        """
```

PR #361, opened 2025-04-15
Detect parameters with duplicate IDs

### LINKED ISSUE #300: ENH: Check if different transformer/line parameters with the same ID are used
The following should raise an error because two different `TransformerParameters` objects with the same ID are used in the same network
```python
tp1 = rlf.TransformerParameters(id="TP", ...)
tp2 = rlf.TransformerParameters(id="TP", ...)

tr1 = rlf.Transformer(parameters=tp1, ...)
tr2 = rlf.Transformer(parameters=tp2, ...)

en = rlf.ElectricalNetwork(transformers=[tr1, tr2], ...)  # <- this should raise an error
```

alihamdan:
This is a proof of concept for detecting duplicate IDs for line parameters in rlfs. It will be applied to transformers and to rlf as well if the impelmentation is deemed good. The implementation is a bit tricky, I am not sure if I covered all edge cases here. Suggestions for improvements or test cases are more than welcome.

Resolves #300

### DIFF
--- a/roseau/load_flow_single/io/tests/test_dict.py
+++ b/roseau/load_flow_single/io/tests/test_dict.py
@@ -50,11 +50,10 @@ def test_to_dict():
         insulator=rlfs.Insulator.PVC,
         section=120,
     )
-    lp2 = rlfs.LineParameters("test", z_line=1, y_shunt=1.1)
 
     geom = LineString([(0.0, 0.0), (0.0, 1.0)])
     line1 = rlfs.Line(id="line1", bus1=source_bus, bus2=load_bus, parameters=lp1, length=10, geometry=geom)
-    line2 = rlfs.Line(id="line2", bus1=source_bus, bus2=load_bus, parameters=lp2, length=10, geometry=geom)
+    line2 = rlfs.Line(id="line2", bus1=source_bus, bus2=load_bus, parameters=lp1, length=10, geometry=geom)
     en = rlfs.ElectricalNetwork(
         buses=[source_bus, load_bus],
         lines=[line1, line2],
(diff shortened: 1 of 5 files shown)


alihamdan:
@benoit9126 when you have time please take a look at this poc (no rush). I'll continue the work when we agree on the implementation.

benoit9126:
It seems to work well :+1: 

A small question regarding the consistency of the code for an implementation in RLF:

In the `network_to_dict` function, an error is raised when params are “different” (using their `__eq__` method). In this PR, you use the `is not` operator in `_add_line_parameters` (`params is not self._line_parameters[params.id]`)

```python
# RLF network_to_dict function
for line in en.lines.values():
  lines.append(line.to_dict(include_results=include_results))
  params_id = line.parameters.id
  if params_id in lines_params_dict and line.parameters != lines_params_dict[params_id]: # <----- "!=" operator
    msg = f"There are multiple line parameters with id {params_id!r}"
    logger.error(msg)
    raise RoseauLoadFlowException(msg=msg, code=RoseauLoadFlowExceptionCode.JSON_LINE_PARAMETERS_DUPLICATES)
```

It could be good to uniformize the comparison of parameters:
  * Or we allow two line parameters instances with the same fields (all of them) to be identical and, then, to be used seamlessly. One of them (no matter which one) will be serialized. When deserializing, a single line parameter object will be created. In this case, we use the `!=` operator (as defined in `LineParameters`), and we need the immutability of the parameters (for the sake of simplicity).
  * Or we consider that two instances are different even if all their fields are equal ⇒ We use the `is` (and `is not`) operator. The immutability of parameters is not required.

I prefer the second option (which is the one you have chosen in this PR). In this case, we maybe don't need anymore to implement the `__eq__` operator in `LineParameters`. The default `__eq__` operator in Python is `is`.

https://docs.python.org/3.13/reference/datamodel.html#object.__eq__
> By default, object implements [__eq__()](https://docs.python.org/3.13/reference/datamodel.html#object.__eq__) by using is, returning [NotImplemented](https://docs.python.org/3.13/library/constants.html#NotImplemented) in the case of a false comparison: True if x is y else NotImplemented.

alihamdan:
Good question, I also prefer identity comparison (with `is`) because it is less error prone when we change the fields. I'll implement your suggestion of using identity comparisons everywhere.

Saelyos (approved):

roseau/load_flow/io/dict.py
```diff
@@ -273,40 +273,30 @@ def network_to_dict(en: "ElectricalNetwork", *, include_results: bool) -> JsonDi
...
     transformers: list[JsonDict] = []
     transformers_params_dict: dict[Id, TransformerParameters] = {}
     for transformer in en.transformers.values():
         transformers.append(transformer.to_dict(include_results=include_results))
         params_id = transformer.parameters.id
-        if params_id in transformers_params_dict and transformer.parameters != transformers_params_dict[params_id]:
-            msg = f"There are multiple transformer parameters with id {params_id!r}"
-            logger.error(msg)
-            raise RoseauLoadFlowException(
-                msg=msg, code=RoseauLoadFlowExceptionCode.JSON_TRANSFORMER_PARAMETERS_DUPLICATES
```
benoit9126, line 292: Idem for `JSON_TRANSFORMER_PARAMETERS_DUPLICATES`

benoit9126 (approved):
:+1: Only remains a small question about error codes

PR #369, opened 2025-04-28
Add results of ground potential to shunt line

alihamdan:
This uses the newly added C++ branch methods to access the potential of the ground port of a line with shunt componenets. This makes the line results accessible independent of the Python ground element. No breaking changes are made to the line results interface but the network JSON format needed to be updated to store the new results.

This work will be essential later when we simplify ground handling in RLF.


roseau/load_flow/models/branches.py
```diff
@@ -175,6 +189,9 @@ def _refresh_results(self) -> None:
         if self._fetch_results:
             self._res_currents = self._cy_element.get_side_currents(self._n1, self._n2)
             self._res_potentials = self._cy_element.get_side_potentials(self._n1, self._n2)
+            if self._has_ground:
+                all_potentials = self._cy_element.get_potentials(self._n)
+                self._res_ground_potential = all_potentials.item(-1)
```
Saelyos, line 194: This is not the fastest way to get results as we are getting twice the potentials. If speed is problematic we can add a C++ methods to get all the 3 potentials of a shunt line.  (this method is not particularly slow, but it can typically get called a lot of times)

Saelyos (commented):
Is there a reason why you added some code to the `Branch` class ? Shouldn't `_has_ground` and `_res_ground_potentials` be only members of `Line` ? 
Otherwise, looks good to me.

roseau/load_flow/models/branches.py
alihamdan: Yeah it would be great if we can ask for the current/potential results of a certain port in C++. Say something like `get_port_potential(i)` where `i` is the index of the port. This could exist on the `Element` class, no need to special case shunt lines. It could be useful for other elements as well that only have one useful port like grounds, prefs and ground connections. The entire RLFS could also use the more efficient methods.

Saelyos: Yes good idea, I can add it in my next PR in `engine`

alihamdan:
> Is there a reason why you added some code to the `Branch` class ? Shouldn't `_has_ground` and `_res_ground_potentials` be only members of `Line` ?

To avoid repeating the methods in the line only to handle this result ( `from_dict`, `_to_dict`, `refresh_results`, etc). I can move it back to the line if you feel strongly about it (like I had it done in the first draft before pushing).

alihamdan:
> Aren't those methods already overridden in lines ? Anyway I was just wondering, I let you decide what is the best.

No all of them. Anyhow, I moved the new logic back to the line as you suggested.

The only thing remaining is the performance of getting this result. Should I merge this PR now and improve the performance later or should I wait for your next engine PR.

roseau/load_flow/models/branches.py
alihamdan: This now uses `self._cy_element.get_port_potential`

PR #370, opened 2025-04-29
Add debuging methods to the solver

Saelyos:
No user facing changes, this PR add some solver methods for debugging purpose.

### DIFF
--- a/roseau/load_flow/_solvers.py
+++ b/roseau/load_flow/_solvers.py
@@ -4,11 +4,12 @@ import warnings
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Generic
 
+import numpy as np
 from typing_extensions import TypeVar
 
 from roseau.load_flow.exceptions import RoseauLoadFlowException, RoseauLoadFlowExceptionCode
 from roseau.load_flow.license import activate_license, get_license
-from roseau.load_flow.typing import FloatArray, FloatMatrix, JsonDict, Solver
+from roseau.load_flow.typing import FloatArray, FloatArrayLike1D, FloatMatrix, JsonDict, Solver
 from roseau.load_flow.utils import find_stack_level
 from roseau.load_flow_engine.cy_engine import (
     CyAbstractNewton,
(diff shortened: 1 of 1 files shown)


alihamdan (approved):

roseau/load_flow/_solvers.py
```diff
@@ -162,9 +163,24 @@ def save_matrix(self, prefix: str) -> None:
         """
         raise NotImplementedError(f"save_matrix() is not implemented for solver {self.name!r}.")
 
-    def current_jacobian(self) -> FloatMatrix:
+    def jacobian(self) -> FloatMatrix:
```
alihamdan, line 166: Would this be better as a property similar to other properties?

roseau/load_flow/_solvers.py
```diff
@@ -162,9 +163,24 @@ def save_matrix(self, prefix: str) -> None:
...
+
+    @variables.setter
+    def variables(self, variables: FloatArrayLike1D) -> None:
+        """Set the independent variables (useful for debugging)."""
+        self._cy_solver.set_variables(np.array(variables, dtype=np.float64))
+
+    @property
+    def residuals(self) -> FloatArray:
+        """Get the residuals of the current iteration (useful for debugging)."""
+        raise NotImplementedError(f"residuals() is not implemented for solver {self.name!r}.")
```
alihamdan, line 183: One more minor thing, since this is a property
```suggestion
        raise NotImplementedError(f"residuals is not implemented for solver {self.name!r}.")
```

roseau/load_flow/_solvers.py
Saelyos: The difference is that this method can be really slow on large networks as it is computing a non-sparse version of the jacobian (with complexity O(n²)). I don't know if it is specific to my debugger, but it tends to compute all the properties when looking at an object so I didn't want to have a slowdown for this. Anyway I have a feeling that properties shouldn't be used for something not immediate that requires heavy computations.

PR #372, opened 2025-04-30
[single] Performance improvements

alihamdan:
Use faster methods to get scalar C++ results and a few minor improvements. This makes hosting capacity calculation from 25% faster for small networks to 300% faster for larger networks.

### DIFF
--- a/roseau/load_flow_single/models/branches.py
+++ b/roseau/load_flow_single/models/branches.py
@@ -1,8 +1,8 @@
 import logging
+import math
 import warnings
 from typing import Self
 
-import numpy as np
 from shapely.geometry.base import BaseGeometry
 from typing_extensions import TypeVar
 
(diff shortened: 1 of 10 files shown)


roseau/load_flow_single/models/branches.py
```diff
@@ -83,7 +83,7 @@ def _check_same_voltage_level(self) -> None:
         if (
             self.bus1._nominal_voltage is not None
             and self.bus2._nominal_voltage is not None
-            and not np.isclose(self.bus1._nominal_voltage, self.bus2._nominal_voltage)
+            and not math.isclose(self.bus1._nominal_voltage, self.bus2._nominal_voltage)
```
alihamdan, line 86: This is minor but math/cmath `isclose` are 60x faster than numpy's and they work with numpy scalars.

roseau/load_flow_single/network.py
```diff
@@ -993,15 +993,13 @@ def _propagate_voltages(self) -> None:
 
     def _get_starting_voltage(self) -> tuple[complex, VoltageSource]:
```
alihamdan, line 994: Simplified as the old logic was needed with multiphase networks only. Here there is no need to create a sorted list.

Issue #346, opened 2025-03-06
BUG: [single] Segfault when adding a line with shunt components to an existing network


alihamdan:
```python
import roseau.load_flow_single as rlfs

bus = rlfs.Bus("Bus")
rlfs.VoltageSource("Source", bus=bus, voltage=20e3)
en = rlfs.ElectricalNetwork.from_element(bus)
bus_new = rlfs.Bus("New Bus")
lp = rlfs.LineParameters("LP with shunt", z_line=0.1 + 0.1j, y_shunt=0.01j)
line_new = rlfs.Line("New Line", bus1=bus, bus2=bus_new, parameters=lp, length=0.1)  # <- segfault here
```

The segfault is triggered when connecting the (internal) ground to the new line.

alihamdan:
This happens because `en._connect_element` gets called before the `_cy_element` of the line is created. So, on this line `self._ground.connect(element._cy_element, [(0, 2)])` the ground is being connected to `None`.

PR #374, opened 2025-05-06
Refactor network and element classes

alihamdan:
- User facing changes:
  - Rename `ElectricalNetwork.from_dgs` to `from_dgs_file` and add `ElectricalNetwork.from_dgs_dict`. The old method is deprecated.
- Internal changes:
  - Refactor common parts of electrical network classes into an abstract base class.
  - Refactor common parts of element classes into an abstract base class.
  - Streamline abstract classes handling
  - Avoid setting `_cy_element` to `None` and prefer deleting the disconnected object to avoid segfaults similar to #346
  - Emit a clear error when mixing rlf and rlfs objects


benoit9126 (approved):

Saelyos (changes requested):
I'm not sure of the use of `self._elements_by_type["bus/..."]` compared to accessing directly `self.buses/...`, but I guess it is easier in some cases

Saelyos (approved):

PR #387, opened 2025-07-15
Stop reformatting arrays in json output

alihamdan:
It is not needed anymore as we use biome to format the json files. This also improves stability of json serialization because arbitrary strings can be used as element IDs which could fool the regular expression.

Also change `dumps` to `dump` for slightly better memory efficiency as it does the writing in chunks.

### DIFF
--- a/roseau/load_flow/utils/mixins.py
+++ b/roseau/load_flow/utils/mixins.py
@@ -171,14 +171,9 @@ class JsonMixin(metaclass=ABCMeta):
             The expanded and resolved path of the written file.
         """
         res = self.to_dict(include_results=include_results)
-        output = json.dumps(res, ensure_ascii=False, indent=2, default=_json_encoder_default)
-        # Collapse multi-line arrays of 2-to-4 elements into single line
-        # e.g complex value represented as [real, imag] or rows of the z_line matrix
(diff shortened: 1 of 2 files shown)


roseau/load_flow/utils/mixins.py
```diff
@@ -171,14 +171,9 @@ def to_json(self, path: StrPath, *, include_results: bool = True) -> Path:
...
-        output = json.dumps(res, ensure_ascii=False, indent=2, default=_json_encoder_default)
-        # Collapse multi-line arrays of 2-to-4 elements into single line
-        # e.g complex value represented as [real, imag] or rows of the z_line matrix
-        output = re.sub(r"\[(?:\s+(\S+,))?(?:\s+?( \S+,))??(?:\s+?( \S+,))??\s+?( \S+)\s+]", r"[\1\2\3\4]", output)
-        if not output.endswith("\n"):
-            output += "\n"
         path = Path(path).expanduser().resolve()
-        path.write_text(output)
+        with path.open("w", encoding="utf-8") as fp:
+            json.dump(res, fp, ensure_ascii=False, indent=2, default=_json_encoder_default)
```
benoit9126, line 176: It is maybe time to use `orjson`?

benoit9126 (approved):

roseau/load_flow/utils/mixins.py
alihamdan: I am planning to add support for orjson later as an optional, used by default if installed, json engine.

PR #392, opened 2025-07-22
Fix disconnecting loads and sources

alihamdan:
The following changes are made:
- Disconnect a ground connection of a load or source when the load or source is disconnected from the network
- Raise a clear error message when accessing any result of a disconnected element
- Add `is_disconnected` property to loads, sources and ground connections to check if the element is disconnected
- In the future, accessing the bus of a disconnected load or source will return the original bus instead of `None`. Use `is_disconnected` to check if the load or source is disconnected.

About changing the bus to no longer be `None` for disconnected elements, I find myself needing to access the bus in many cases mainly to connect another element to it. The new behavior works well as it guarantees a clear error message when the user tries to access the results of a disconnected element while keeping access to its bus.

### DIFF
--- a/roseau/load_flow/models/connectables.py
+++ b/roseau/load_flow/models/connectables.py
@@ -77,11 +77,7 @@ class AbstractConnectable(AbstractTerminal[_CyE_co], ABC):
         self._res_currents: ComplexArray | None = None
 
     def __repr__(self) -> str:
-        args = [
-            f"id={self.id!r}",
-            f"bus={repr(self.bus.id) if self.bus is not None else '<disconnected>'}",
-            f"phases={self.phases!r}",
-        ]
+        args = [f"id={self.id!r}", f"bus={self.bus.id!r}", f"phases={self.phases!r}"]
         if self._connect_neutral is not None:
             args.append(f"connect_neutral={self._connect_neutral!r}")
         side = f"-{self._side_value}" if self._side_value is not None else ""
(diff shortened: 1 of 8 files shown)


roseau/load_flow/models/connectables.py
```diff
@@ -77,11 +77,7 @@ def __init__(
         self._res_currents: ComplexArray | None = None
 
     def __repr__(self) -> str:
-        args = [
-            f"id={self.id!r}",
-            f"bus={repr(self.bus.id) if self.bus is not None else '<disconnected>'}",
-            f"phases={self.phases!r}",
-        ]
+        args = [f"id={self.id!r}", f"bus={self.bus.id!r}", f"phases={self.phases!r}"]
```
benoit9126, line 80: The `self.bus.id` will fail is the load is disconnected as long as the `bus` property retrieves `None` in this case.

roseau/load_flow_single/models/connectables.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, id: Id, bus: Bus, *, n: int, side: Side | None = None) -> Non
         self._res_current: complex | None = None
 
     def __repr__(self) -> str:
-        args = [f"id={self.id!r}", f"bus={repr(self.bus.id) if self.bus is not None else '<disconnected>'}"]
+        args = [f"id={self.id!r}", f"bus={self.bus.id!r}"]
```
benoit9126, line 46: idem

benoit9126 (changes requested):
Two small changes are required as long as the `bus` property may retrieve `None`

benoit9126 (approved):

PR #399, opened 2025-07-31
Add popup and search to interactive map plots

alihamdan:
Also fix a bug introduced in #395 when there are lines with heterogeneous parameters (like a different neutral section) where folium receives a numpy array instead of a list and raises.


roseau/load_flow/plotting.py
```diff
@@ -371,7 +387,7 @@ def _scalar_if_unique(value):
             unique = np.unique(value)
             if unique.size == 1:
                 return unique.item()
-            return value
+            return value.tolist()
```
alihamdan, line 390: This fixes the bug mentioned in the PR description

alihamdan:
A preview of the added features:
image

Saelyos (approved):

PR #411, opened 2025-09-25
[single] Performance improvements of setters

Saelyos:
Use faster methods to update the parameters of loads, lines and voltage sources. This makes the updating about 10%~15% faster.


alihamdan (commented):
So here you only change the "update" methods but still require arrays when creating the cython elements? Would it make sense to allow passing scalars when creating the cython elements as well or for example when setting initial potentials of cython voltage sources? or perhaps is it not worth it in terms of performance.

alihamdan:
On a second thought, I think at least `initialize_potentials` would be good to improve as it is called a lot when propagating the voltages of the network.

PR #418, opened 2025-12-03
Improve interactive map plots

alihamdan:
Many improvements to the styles of interactive map plots:

  - Different voltage levels are represented with different marker sizes and line widths. Nominal voltages are used to
    determine the voltage levels automatically. If nominal voltages are not available, they are inferred from the
    transformers and sources
  - Underground lines are dashed while other lines are solid including lines with unknown line type
  - Transformers are now represented with a square icon with a divider in the middle and with both HV and LV buses
    information in the tooltip and popup
  - Sourcers are represented with a bigger square icon
  - If the new parameter `fit_bounds` is set to `True` (default), the bounds of the map will be automatically adjusted
    using `folium.FitOverlays` to fit the network elements.

  Note that the markers of buses, transformers and sources now use the `folium.DivIcon` icon which means the style
  function must now return a style dictionary of the form `{'html': ...}` for these elements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,6 @@ on:
       - "doc/**"
       - ".vscode/**"
 
-concurrency:
-  group: ci-${{ github.ref }}
-  cancel-in-progress: false
-
 env:
   CI: true
   UV_SYSTEM_PYTHON: 1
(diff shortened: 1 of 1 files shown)


Saelyos (approved):
Thanks for the modification, LGTM now

benoit9126 (approved):

Issue #435, opened 2026-01-07
BUG: `rlf.LineParameters.to_sym()` should perform Kron's reduction for 4-wire lines


alihamdan:
When converting to symmetrical components we eliminate the neutral so we need to perform a correction for the impedance using Kron's reduction.

This works fine in rlfs because rlfs is only concerned by the positive-sequence parameters but this means that zero-sequence parameters for 4-wire lines are not correct.

Issue #422, opened 2025-12-10
ENH: Add LV cables to the catalogue


alihamdan:
Example: https://www.nexans.ma/.rest/eservice/dam/v1/file/246839/Catalogue%20BT_Nexans.pdf

Issue #298, opened 2024-12-05
ENH: A function to plot the voltage profile of a network


alihamdan:
The idea is to have a graphic representation of the voltage levels in the whole network. We start from an initial node and walk the graph of the network from this node. The user will be able to choose the initial node, by default we choose a bus having a voltage source. The resulting graphic would show the voltage level on the y-axis in p.u. (or in percent) and the distance on the x-axis (in km).

PR #438, opened 2026-01-21
Version 0.14.0

benoit9126:
This version 0.14.0 has some major improvements! Please note that some deprecated functions, packages and attributes have been removed in this version.

**Changelog**

- #435 #436 Add `rlf.converters.kron_reduction` function to perform Kron reduction on any nxn matrix reducing it to an (n-1)x(n-1) matrix by eliminating the last row and column. Also add `eliminate_neutral` parameter to   `rlf.LineParameters.to_sym` to optionally eliminate the neutral conductor using Kron's reduction when converting line parameters to symmetrical components.

- #434 #422 Add LV line parameters to the catalogue (e.g. `U_AL_3x150+70`). You can get all added LV lines with `rlf.LineParameters.get_catalogue(r"^[UOT]_[A-Z]+_3x(\d+)\+(\d+(?:\.\d+)?)$")`.

- #429 #298 Add `rlf.plotting.voltage_profile` function to plot the voltage profile of a network. The function returns an object with `plot_matplotlib()` and `plot_plotly()` methods to create plots using Matplotlib or Plotly respectively.

- #427 Add missing transformer tap to the edge data in the graph generated by `ElectricalNetwork.to_graph`.

- #426 Add `rlf.plotting.plot_results_interactive_map` for plotting a network with load flow results on an
  interactive map. Buses are colored according to their voltage levels and lines are colored according to their loading
  levels.

- #425 Enforce the following deprecations:

  - The `side` parameter of `rlf.GroundConnection`, `rlf.plotting.plot_voltage_phasors` and
    `rlf.plotting.plot_symmetrical_voltages` is removed. Pass a branch side directly instead (e.g.
    `rlf.GroundConnection(ground=ground, element=transformer.side_hv)`).
  - The `potentials` attribute and constructor parameter of `Bus` is removed. Use `initial_potentials` instead.
  - The `bus` of a disconnected load or source no longer returns `None`. Use the new `is_disconnected` property to check
    if the load or source is disconnected.
  - The `phase` parameter of `rlf.PotentialRef` is removed. Use `phases` instead.
  - The `res_voltages_hv` and `res_voltages_lv` attributes of `rlf.Transformer` are removed. Use `<side>.res_voltages` instead.
  - Deprecated names in `rlf.constants`, `rlf.types` and `rlf.sym` modules are removed.
  - Deprecated modules in the `rlf.utils` package are removed. This package is for internal use only and should not be considered stable.

- #421 Improve initial voltage guesses in the Newton algorithm for networks involving delta connected sources.

- #419 #420 Require Python 3.12 or newer and bump the minimum supported versions of some dependencies per [SPEC-0](https://scientific-python.org/specs/spec-0000/).

- #418 Many improvements to the styles of interactive map plots:

  - Different voltage levels are represented with different marker sizes and line widths. Nominal voltages are used to
    determine the voltage levels automatically. If nominal voltages are not available, they are inferred from the
    transformers and sources
  - Underground lines are dashed while other lines are solid including lines with unknown line type
  - Transformers are now represented with a square icon with a divider in the middle and with both HV and LV buses
    information in the tooltip and popup
  - Sourcers are represented with a bigger square icon
  - If the new parameter `fit_bounds` is set to `True` (default), the bounds of the map will be automatically adjusted
    using `folium.FitOverlays` to fit the network elements.

  Note that the markers of buses, transformers and sources now use the `folium.DivIcon` icon which means the style
  function must now return a style dictionary of the form `{'html': ...}` for these elements.

- #413 Restore bus and layers to the layer control in the interactive map plot. This was broken in version 0.13.0.

- #404 Fix filtering catalogues using compiled regular expressions with newer versions of pandas.

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -44,7 +44,7 @@ og:description: See what's new in the latest release of Roseau Load Flow !
   - The `side` parameter of `rlf.GroundConnection`, `rlf.plotting.plot_voltage_phasors` and
     `rlf.plotting.plot_symmetrical_voltages` is removed. Pass a branch side directly instead (e.g.
     `rlf.GroundConnection(ground=ground, element=transformer.side_hv)`).
-  - The `potentials` attribute and and constructor parameter of `Bus` is removed. Use `initial_potentials` instead.
+  - The `potentials` attribute and constructor parameter of `Bus` is removed. Use `initial_potentials` instead.
   - The `bus` of a disconnected load or source no longer returns `None`. Use the new `is_disconnected` property to check
     if the load or source is disconnected.
   - The `phase` parameter of `rlf.PotentialRef` is removed. Use `phases` instead.
--- a/roseau/load_flow/tests/test_electrical_network.py
+++ b/roseau/load_flow/tests/test_electrical_network.py
@@ -737,7 +737,7 @@ def test_buses_voltages(small_network_with_results):
         pd.DataFrame.from_records(voltage_records)
         .astype(
             {
-                "bus_id": str,
+                "bus_id": object,
                 "phase": VoltagePhaseDtype,
                 "voltage": complex,
                 "voltage_level": float,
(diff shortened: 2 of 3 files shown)


alihamdan (approved):
Let’s go 🚀

PR #449, opened 2026-03-04
Fix I0 of some transformers in the catalogue

alihamdan:
The no-load currents (I0) of some of the most recent transformers in the catalogue were off by a factor of 100. The values in the manufacturers catalogue are in % while the values we save are in pu. This PR fixes these currents for affected transformers.

Note that this considerably improves the reactive power losses in these transformers, up to a factor of 2 when the transformer is fully loaded:
image


alihamdan:
@Saelyos you'll need to regenerate your transformer interpolation parameters with this fix applied.

Saelyos (approved):
Good catch @alihamdan 
Yes I'll regenerate the interpolation

benoit9126 (approved):

PR #451, opened 2026-04-23
Fix switches in plots, ty, and uv build warning

alihamdan:
- Fix missing switches with linestring geometries in the interactive map plots
- Fix a uv warning about a new version of uv build
- Improve compatibility of `rlf.Q_` with ty with latest pint versions
- Update the dependencies and pre-commit hooks


alihamdan:
Here is the plots before:
image

And after with the switches:
image

benoit9126 (approved):

alihamdan:
@benoit9126 we are ready for the release

PR #466, opened 2026-06-23
Minor cleanups and micro-optimizations

alihamdan:
I had claude audit the code to make some improvements

----------------------
- Replace scalar np.isclose with math.isclose/cmath.isclose (with abs_tol=1e-8 for zero checks to preserve np.isclose default behavior)
- Use np.asarray instead of np.array to avoid copying already-array inputs
- Access private float attributes directly in _to_dict instead of constructing Q_ objects just to read back .magnitude


alihamdan:
This is minor so I am going to merge directly

PR #472, opened 2026-07-08
Improve interactive map plots and fix highlighting

alihamdan:
- Fix highlighting of buses and transformers on hover (broken since the use of div icon in the marker)
- Reduce the size of the markers for better visibility of networks with many elements as is the case of most bdgd feeders
- Use `"CartoDB Positron"` as default tiles for better visibility of the elements. Pass `tiles="OpenStreetMap"` to use OpenStreetMap tiles.

Also replaced manual centering of buses on their x,y coordinates hack by automatic css centering using `position: absolute` and `transform: translate(-50%, -50%)` which works with any marker size.

### DIFF
--- a/roseau/load_flow/plotting.py
+++ b/roseau/load_flow/plotting.py
@@ -3,7 +3,7 @@
 import cmath
 import dataclasses
 import math
-from collections.abc import Callable, Iterable, Mapping
+from collections.abc import Callable, Container, Iterable, Mapping
 from typing import TYPE_CHECKING, Any, Literal, Self, TypedDict
 
 import geopandas as gpd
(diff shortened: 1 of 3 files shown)


roseau/load_flow/plotting.py
```diff
@@ -478,14 +527,32 @@ def internal_highlight_function(feature):
...
+    # (an `{"html": ...}` dict) by rebuilding its icon.
+    root = m.get_root()
+    assert isinstance(root, folium.Figure)
+    folium.Element(
+        "L.Marker.include({"
+        "  setStyle: function (style) {"
+        "    this.setIcon(L.divIcon(Object.assign({}, this.options.icon.options, style)));"
+        "  }"
+        "});"
+    ).add_to(root.script)
```
alihamdan, line 545: Explanation and fix by Claude

Issue #446, opened 2026-02-20
ENH: Add total bus power to the voltage profile and map plots


alihamdan:
Display the sum of the powers of loads and sources of each bus in the tooltips and pop-ups of buses.

PR #479, opened 2026-07-13
Version 0.15.0

benoit9126:
Version 0.15.0 🎉

> [!IMPORTANT]  
> The public license key can now be used with networks containing up to 50 buses. It was previously limited to 10 buses.

- #475 #446 Show more information when hovering over elements in the voltage profile and interactive
  map plots. The hover box of buses now shows the net (load - source) active and reactive powers consumed by the bus.
  Hovering over elements in the voltage profile plots now shows more relevant information about the element.
- #474 Wrap long element IDs in the hover box of the voltage profile plots.
- #473 Add support for custom color per element in the interactive map plots by accepting a callback as the
  `style_color` argument. The callable accepts two positional arguments: the element type and the element ID, and
  returns a color string or `None` to use the default color for that element.
- #472 Many improvements and fixes for the interactive map plots.
  - Fix highlighting of buses and transformers on hover
  - Reduce the size of the markers for better visibility of networks with many elements
  - Use `"CartoDB Positron"` as default tiles for better visibility of the elements. Pass `tiles="OpenStreetMap"` to use
    OpenStreetMap tiles.
- #471 Include switch phases in the interactive map plots.
- #467 Avoid unnecessary copies when deserializing networks from JSON or dict and remove `from_dict` methods
  from element classes, these methods did not work correctly because they depend on other elements.
- #466 Avoid unnecessary copies of already array inputs.
- #465 Make results access on elements between 2x and 3x faster by making the `Q_` class lazy.
- #464 Support compact JSON output with `indent=False` and improve performance by using `orjson` if installed.
- #462 Add `name` attribute to `ElectricalNetwork` to store the name of the network.
- #460 Modify the load flow convergence check to use residuals and tolerance comparison. This fixes the case of
  convergence at exactly the number of iterations specified by the `max_iterations` parameter.
- #459 Add support for three-phase transformers with "untrue" vector groups
- #458 Drop support for the experimental Python 3.13t (free-threaded) following end of support in cibuildwheel
- #457 Fix a rare bug where modifying the tap of an inverted single-phase transformer (vector groups "Ii6" and
  "Iii6") after its creation would incorrectly invert back its orientation.
- #456 Support single-phase transformer parameters conversion from OpenDSS.
- #454 Fix conversion of transformer parameters from OpenDSS when `loadloss` is non-zero.


alihamdan (approved):
Let's go

PR #492, opened 2026-08-21
Cache pint registry for faster import time

alihamdan:
pint supports [caching](https://pint.readthedocs.io/en/develop/advanced/performance.html#speed-up-registry-instantiation) parsed unit registry on disk for faster instantiation. This cuts ~150 ms from our import time on a warm run which is currently very slow.

Also updated the minimum pint version to the one that supports numpy 2 which we require

### DIFF
--- a/doc/Changelog.md
+++ b/doc/Changelog.md
@@ -19,6 +19,10 @@ og:description: See what's new in the latest release of Roseau Load Flow !
 
 # Changelog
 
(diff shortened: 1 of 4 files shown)


benoit9126 (approved):

alihamdan:
Forgot to mention, the `%` unit is now natively supported by pint so I removed the preprocessor

